PR #16, opened 2023-02-11
Added notebook for task 1 (compute returns)

FirstSingularity:
The notebook contains outlier (>3 std away) trades between different cryptocurrencies and USD. I provided HTML so the notebook and its results could be viewed without running it.

Things I am unsure about:
- Should I divide the price difference by the time difference?
- Should each pair (crypto -> usd) be normalized independently or as a whole?
- Is there anywhere I should place the output files to observe later, or should I keep them locally?


sorrentum_sandbox/projects/SorrIssue1_Predict_large_asset_movements_with_NLP/task_1_compute_returns.ipynb
```diff
@@ -0,0 +1,494 @@
...
+      "A value is trying to be set on a copy of a slice from a DataFrame.\n",
+      "Try using .loc[row_indexer,col_indexer] = value instead\n",
+      "\n",
+      "See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy\n",
+      "  frame['normalized_close_change'] = (frame['close_change'] - frame['close_change'].mean()) / frame['close_change'].std()\n"
+     ]
+    }
+   ],
+   "source": [
+    "split_dfs = {i: None for i in df['currency_pair'].unique()}\n",
```
gpsaggese, line 120: In pandas you don't want to do loops, but want to do things vectorially otherwise you can't scale up to the amount of data we need to process.

The groupby is fine to split by currency, and then on each df you should do:
```
df["rets"] = df["close"].pct_change()
```

An even better approach is to do a multi-index df where the outermost loop is the currency.
Then you don't even have to do a loop on the currency. It's done in one shot.

It would be good to resample to 1T adding nans to make sure the grid is regular so you don't have to worry about time gaps.
https://pandas.pydata.org/docs/user_guide/advanced.html

I've also written a bunch of tutorials from the PyData book here

https://github.com/gpsaggese/umd_data605/tree/main/tutorials/tutorial_pandas

sorrentum_sandbox/projects/SorrIssue1_Predict_large_asset_movements_with_NLP/task_1_compute_returns.ipynb
```diff
@@ -0,0 +1,494 @@
...
+    "# 2\n",
+    "for pair, frame in split_dfs.items():\n",
+    "    close_price = frame['close']\n",
+    "    time_diff = frame['timestamp']\n",
+    "    close_change = [None] + [(close_price[i] - close_price[i - 1]) / (time_diff[i] - time_diff[i - 1]) for i in range(1, len(close_price))]\n",
+    "    frame['close_change'] = close_change\n",
+    "\n",
+    "# 3\n",
+    "for pair, frame in split_dfs.items():\n",
+    "    frame['normalized_close_change'] = (frame['close_change'] - frame['close_change'].mean()) / frame['close_change'].std()"
```
gpsaggese, line 135: It's ok in first approx, but this is a non-causal transformation that it's always dangerous, since the outliers depend on the future.

The best approach is to do a rolling mean / std
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html

sorrentum_sandbox/projects/SorrIssue1_Predict_large_asset_movements_with_NLP/task_1_compute_returns.ipynb
```diff
@@ -0,0 +1,494 @@
...
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "data": {
+      "text/plain": "<Figure size 864x432 with 1 Axes>",
+      "image/png": "iVBORw0KGgoAAAANSUhEUgAAAugAAAGECAYAAABkhNUsAAAAOXRFWHRTb2Z0d2FyZQBNYXRwbG90bGliIHZlcnNpb24zLjMuNCwgaHR0cHM6Ly9tYXRwbG90bGliLm9yZy8QVMy6AAAACXBIWXMAAAsTAAALEwEAmpwYAABAKUlEQVR4nO3de1iUZeL/8c/ADKhBS9iMKJ1Ns+ygK1ZmQbUJKqDpamuSdtivpy0z9xsuIGFqmhGZWR7a6upg7ZZrK5gLuGWhu2Gbuh3W1qyfeSY5mggqM8M8vz+8nK+EESDDPMj7dV1eMvfz3HMf5mb48HDPjMUwDEMAAAAATCHA3x0AAAAA8H8I6AAAAICJENABAAAAEyGgAwAAACZCQAcAAABMhIAOAAAAmAgBHQAAADARq787AAA4c1dccYV69uypgIC6112WLFmi1157TZs3b5Yk7dy5U5GRkerQoYMk6Z133tGECROUlJSkwYMHe+vt379fiYmJ+uyzzxpsNyUlRR9//LHCw8NlGIacTqcGDhyo1NRUWa1WPf/88zp06JAyMjLq1EtISNBjjz2mG264QSUlJZo/f7527twpSerQoYMmTZqkO+64Q5J0++23y2azqUOHDjIMQ4ZhaOjQoZowYYKsVn6MATj78MwGAGeJ119/XeHh4fXK09PTvV/ffvvtysrK0jXXXNNi7d5333367W9/K0mqqanRmDFjlJubq2HDhjWqfnp6um666SYtWrRIkvT//t//0913361LL71U3bt3l6Q6fT569Kge
```
gpsaggese, line 425: I would start from even higher, like 4 or 5 std dev. Then pls post the resulting CSV in the gdrive like "asset,timestamp,normalized_ret,ret"

The timestamp should be using ET timezone since news are typically reported with that time.

Then we need to go through the file and start looking at what is the corresponding news.

gpsaggese (commented):
Good work. Feedback below. Let's update the notebook as suggested and then we can do another PR.

Issue #19, opened 2023-02-12
Understand large movements due to news


gpsaggese:
Given the output of #18 and #17 we want to understand what moves price.

The output is a document with a list of large movements and information in what likely happen.

Why do they happen? Google around to try to build an understanding of what moves each coin
Is it a market-wide movement, or is it specific to each coin?

This will inform successive machine learning models

DanilYachmenev:
probably obsolete, moving to P1 for now

PR #31, opened 2023-02-13
Create task_1_compute_returns.ipynb

FirstSingularity:
The notebook contains outlier (>5 std away) trades between different cryptocurrencies and USD. This is the second pull request for this task, since I moved from forking to branching. 

Things I am unsure about:
1. What size should the window be for rolling mean/std?


sorrentum_sandbox/projects/SorrIssue1_Predict_large_asset_movements_with_NLP/task_1_compute_returns.ipynb
```diff
@@ -0,0 +1,529 @@
...
+     },
+     "execution_count": 77,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "df['pct_change'] = df.groupby(level='currency_pair')['close'].pct_change()\n",
```
gpsaggese, line 168: you should cache this in a `ret` variable so the code is cleaner and faster (since you don't recompute the some stuff multiple times)

sorrentum_sandbox/projects/SorrIssue1_Predict_large_asset_movements_with_NLP/task_1_compute_returns.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "df['norm_pct_change'] = (df.groupby(level='currency_pair')['pct_change'].rolling(ROLLING_WINDOW).mean() / df.groupby(level='currency_pair')['pct_change'].rolling(ROLLING_WINDOW).std()).droplevel(0)\n",
+    "df.head()"
```
gpsaggese, line 170: Try to make the cells idempotent by assigning values to different vars (even with horrible names like df1, df2, ...).
In this way you can run the notebook from the middle without carrying over state across cells

sorrentum_sandbox/projects/SorrIssue1_Predict_large_asset_movements_with_NLP/task_1_compute_returns.ipynb
```diff
@@ -0,0 +1,529 @@
...
...u5599lm5XC49/vjjhGoAgGWSnAn8JQTABbFNqJ48ebLOnDmjVatWqbS01OpyAAAAgLDZ5u4f7e3tqqio0NSpU5Wbm2t1OQAAAEDYbBOqn3vuOR0+fFivvvqqXnzxRavLAQAAAMIW9cs/Ojo6VFxcrHXr1mnkyJGSpIaGBtXW1ioQCKisrEwlJSVasWJFtEsBAAAAoiKqoXr37t2qqKhQS0tLaF1ra6uqq6tVX18vl8ul4uJijR8/XllZWabmGjEi1WS1/RfOJ8QRW/TEfqLRk6Sk8A5hkRrncDgsmTea24yHcVb3JdL/tjl+2RN9sR+79SSqobqurk6VlZVauHBhaF1TU5NycnKUnp4uScrPz1djY6Pmzp1raq5jxzoUDBqmttEfbneafL4TMZ8XPaMn9hONnrjdafL7A2GNjdQ4wzAiur0LHWfl3HYeZ3VfIvlvm+OXPdEX+7GqJwkJjh5P5EY1VFdVVZ2z7ujRo3K73aFlj8ej5ubmaJYBAAAARFXMb6kXDAZDf6qTzp5h+ONlAAAGCh4SA+B7MQ/VmZmZ2rVrV2jZ5/PJ4/HEugwAAEzjITEAvhfzUJ2bm6uamhq1tbUpJSVFmzZt0rJly2JdBgCbGjY8Rcku2zyXCoiIcM9o+wPBGFQDIBpi/s6VkZGh8vJylZaWyu/3q6ioSNnZ2bEuA4BNJbucnPlD3OGMNhD/YhKqt2zZ0m3Z6/XK6/XGYmoAAAAg6mzzREUAAABgoCJUAwAAACbxaSAAURfuh7QAABioCNUAoi7cD2lJfFALADAwcfkHAAAAYBKhGgAAADCJUA0AAACYRKgGAAAATCJUAwAAACYRqgEAAACTCNUAAACASYRqAAAAwCRCNQAAAGASoRoAAAAwiVANAAAAmESoBgAAAEwiVAMAAAAmEaoBAAAAk5xWFwBg4Bo2PEXJLg4jAADwbgjgHBcSlhesea/PMSvn55ktCRgU/IGg3O60Psed7gqo/XhnDCoCEC5CNYBzJLuchGXAAknOBH73gAGKa6oBAAAAkwjVAAAAgEmEagAAAMAkQjUAAABgEh9UBABggAn3LiESdwoBYoVQDQDAABPuXUIk7hQCxAqhGhhEeFgLAADRwbsrMIhw/2kAAKKDDyoCAAAAJhGqAQAAAJMI1QAAAIBJXFMNAEAcC/f2e9x6DzCHUA0AQBwL9/Z7fEAZMIfL
```
gpsaggese, line 472: you can use an aux var also here

```
df_abs = ...
df_extreme_vals = df_abs[df_abs[...] > sigma] 
...
```

gpsaggese, line 486: Once the extremes are "fixed" due to volatility, you can also take #19 and start doing some research.

You are moving forward. We will assign other bugs to other teammates

gpsaggese (commented):
Much better. This is quant pandas. Nice job with the multi-index.

You should use a longer window. 4 samples is not long enough. In fact you are getting too much mass of the PDF outside 5 sigmas (the extreme events should be very uncommon).

Estimating volatility is a complex problem, but I'd use 4 days worth of samples so 4 * 3600 * 24.

Let's make the small changes and we can merge. 

Well done

sorrentum_sandbox/projects/SorrIssue1_Predict_large_asset_movements_with_NLP/task_1_compute_returns.ipynb
FirstSingularity: If I make a dedicated df_abs, that would change the normalized_ret to abs(normalized_ret) unless I make an entirely different column. Would it be better to skip df_abs and go straight to df_extreme_values?

FirstSingularity:
I have the resampling set to every minute ('T') rather than every second ('S'). Should I change to seconds and make the window 4 * 3600 * 24, or should I keep it as minutes and make the window 4 * 60 * 24?

gpsaggese:
It doesn't make a difference. I would keep minutes for simplicity and just add a line like:

```
window_length_in_min = str(60 * 24) + "T"
...rolling(samples=window_length_in_min
```

FirstSingularity:
I tried changing up the window size and sigma, but I'm still getting too much volume in the output. For a window of 4 * 24 * 60 and 5 sigma, the output contains ~90k rows. For the same window and 8 sigma, the output is ~20k rows; for a window of 7 days and 5 sigma, the output is ~100k rows. 

The way the normalized returns are being calculated is
```
ret = df1.groupby(level='currency_pair')['close'].pct_change().rename('ret')
grouped_rolling_ret = ret.groupby(level='currency_pair').rolling(ROLLING_WINDOW)
normalized_ret = ((ret - grouped_rolling_ret.mean().droplevel(0)) / grouped_rolling_ret.std().droplevel(0)).rename('normalized_ret')
```

gpsaggese:
The code is fine.

The resampling is a rolling one, so the window size doesn't affect the length of the output (just a bit through the burn-in to get enough samples). 

The fact that there are lots of results is due to:
2) non-gaussianity of returns (so 5 sigma events are much more frequent and what a gaussian would imply)
3) Likely the same news generates lots of spikes across coins since they are all highly correlated

Another approach is to resample the data to 5 mins to look for longer spikes but I would stick to short spikes.

Overall that's not a problem, I would rank the returns by absolute magnitude and pick the first 100.

Then we can go through them and see what's the news that generates them.
The goal is to automate this process, but we need to understand the system before automating.

We can talk today.

FirstSingularity:
I won't be able to make the meeting today, but that all makes sense to me; I made all the changes and will commit soon. Which Google Drive dir should I upload the outputs to?

gpsaggese:
https://github.com/sorrentum/sorrentum/discussions/35#discussioncomment-4974408

PR #53, opened 2023-03-08
Issue 19: Ravenpack 2021 Crypto NLP

FirstSingularity:
Contains work towards matching news with outlier trades for 2021.

1. Import outlier data
2. Import RP data (with specific features)
3. Change RP data timezone to America/New York
4. Query RP data to find articles close in timestamp to each outlier event

Things I am unsure about:
1. Is querying for relevant RP data the most efficient way to do it?
2. What range of time should I look for relevant articles in?


FirstSingularity:
@okritvik I'll try to finish up pairing the articles with outliers, since that seems like a one-person job and then I'll upload that data to the shared drive.

FirstSingularity:
@sibiakash11 the final block in the commit displays all the query results from the RP data (articles) from one minute before a specific outlier trade. So far, it looks like news about other cryptocurrencies influence the coin that the outlier trade is based on.

okritvik:
> @okritvik I'll try to finish up pairing the articles with outliers, since that seems like a one-person job and then I'll upload that data to the shared drive.

Sounds good. I will search up DL models for embedding.

okritvik:
@FirstSingularity I looked at the code and it's great! Would it be possible for you to write some comments in the code about what is actually happening in that block? It would be easy for others who would like to work on that piece or take it as a reference in the future.

Issue #66, opened 2023-03-22
Create an external adapter to get TWAP/ VWAP prices for DaoSwap


tamriq:
As we discussed, we want to have three price sources for DaoSwap:
1) Chainlink Price Feed (already implemented). In fact, the price in these feeds is typically derived from multiple sources, including different exchanges and data providers.
They do not directly use TWAP or VWAP as part of aggregation. As far as I understand, they do use a sophisticated aggregation mechanism that takes into account various factors, including the volume, liquidity, and quality of data sources. 
2) TWAP (not implemented).
3) VWAP (not implemented). 

While Chainlink Price Feeds do not explicitly provide TWAP or VWAP data, we should create a custom solution with [chainlink external adapters](https://docs.chain.link/chainlink-nodes/external-adapters/external-adapters/). About external adapters: 
 - An external adapter for Chainlink is a small program or service (in fact it can be a python web-service written in flask) that helps Chainlink nodes do extra tasks or get data from sources that they can't handle by default. External adapters are like helpers that extend what Chainlink nodes can do. They allow the node to connect to more complex systems, access different types of data, or perform custom calculations.
 - When a smart contract needs some data or computation that a Chainlink node can't provide directly, it can ask for the help of an external adapter. The node sends the request to the external adapter, which does the job and sends the result back to the node. The node then provides the result to the smart contract.

So in fact here we need to:
1) Choose a data source. We need to find a reliable data source that provides TWAP or VWAP data. Many popular cryptocurrency exchanges and data providers offer API access to historical or real-time TWAP and VWAP values. 
2) Create an external adapter. The adapter should be able to receive requests from a Chainlink node, query the API for TWAP or VWAP data, and return the results in a format compatible with Chainlink nodes.  

Chainlink nodes send requests to external adapters using the POST method by default, so overall we just need to create small Flask web app with two `get_twap` and `get_vwap` POST methods.

tamriq:
FYI @gpsaggese

gpsaggese:
Specs are very clear and this looks like a fun project!

1) Let's build the simplest skateboard possible (e.g., get the data from chainlink and accumulate the TWAP and serve it).

2) FYI we could use our own interface to Binance to use the prices (e.g., @dremdem and @jsmerix) but in general we want to get DaoSwap as independent from our stuff as possible, so we would just copy-paste code rather than relying on our Airflow infra.

[...]

PR #70, opened 2023-03-23
Issue66: Add TWAP/VWAP external adapter

DanilYachmenev:
Task #66 

We need to test it, but should work
at least the API query replies:
https://api.coingecko.com/api/v3/coins/bitcoin/market_chart/range?vs_currency=eth&from=1640995200&to=1648060799&interval=minutely


defi/dao_swap/twap_vwap_adapter.py
```diff
@@ -0,0 +1,77 @@
...
+import json
+
+import numpy as np
+import requests
+from flask import Flask, jsonify, request
+
+app = Flask(__name__)
+
+# TODO(Dan): Replace with a company API. Now a dummy is used.
+_API_KEY = "bef8f874d2ad2e08d1e0eb91bad9c53399b8ca55016ba01101f4eeb855abf70f"
```
DanilYachmenev, line 16: Note that `_API_KEY` should be changed, now I've just created 1 with 10-min-mail to test functionality

defi/dao_swap/twap_vwap_adapter.py
```diff
@@ -0,0 +1,80 @@
...
+    volumes = [
+        float(entry["volumefrom"])
+        for entry in price_data
+        if entry["time"] >= start_time
+    ]
+    return prices, volumes
+
+
+@app.route("/get_twap", methods=["POST"])
+def get_twap():
```
DanilYachmenev, line 48: do we need type hints in the app?

defi/dao_swap/twap_vwap_adapter.py
gpsaggese: Can we use a value passed through env var? Even a dummy key will trigger alarms when we release.
@tamriq knows how to do it

defi/dao_swap/twap_vwap_adapter.py
```diff
@@ -0,0 +1,80 @@
...
+import requests
+from flask import Flask, jsonify, request
+
+app = Flask(__name__)
+
+# TODO(Dan): Replace with a company API. Now a dummy is used.
+_API_KEY = "bef8f874d2ad2e08d1e0eb91bad9c53399b8ca55016ba01101f4eeb855abf70f"
+
+
+def _get_price_volume_data():
```
gpsaggese, line 19: Any reason for not using type hints?

gpsaggese (commented):
LGTM

defi/dao_swap/twap_vwap_adapter.py
tamriq: I think we do

tamriq (commented):
Looks goods, let's hide API key to `os.environ` and return price in WEI.

defi/dao_swap/twap_vwap_adapter.py
DanilYachmenev: Added simple ones, linter swears and asks just for `Any`

defi/dao_swap/twap_vwap_adapter.py
DanilYachmenev: wanted to know if they are needed in apps, since they are not in tasks

defi/dao_swap/twap_vwap_adapter.py
```diff
@@ -0,0 +1,80 @@
...
+    Query price and volume data from the CryptoCompare API.
+    """
+    # Get parameters from the Chainlink node request.
+    symbol = request.json["symbol"]
+    start_time = request.json["start_time"]
+    end_time = request.json["end_time"]
+    request.json["time_interval"]
+    # Query the CryptoCompare API for price data within the specified time range.
+    response = requests.get(
+        f"https://min-api.cryptocompare.com/data/v2/histohour?fsym={symbol}&tsym=USD&limit=2000&aggregate=1&toTs={end_time}&api_key={_API_KEY}"
```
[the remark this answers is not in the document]
DanilYachmenev: Switched ro CoinGecko and it made everything even easier - no need for API KEY for now

However, free version has 10 - 30 calls per min limit and rn I'm thinking that it is not enough
still, lets decide if its truly insufficient and look for alternatives if so

because pro version turned out to be terribly expensive:
Снимок экрана 2023-03-24 в 01 55 40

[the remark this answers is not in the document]
DanilYachmenev: hmm, though it still seems to be one of the best available options among other advanced plans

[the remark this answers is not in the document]
tamriq: 10 - 30 calls per min is in fact not enough, but for now it's fine

tamriq (approved):

PR #72, opened 2023-03-26
Add stages to DaoSwap contract to limit creation of new orders

KangmingL:
Add stages to limit the creation of new sell/buy orders when OnSwapTime is in processing.

### DIFF
--- a/defi/dao_swap/contracts/DaoSwap.sol
+++ b/defi/dao_swap/contracts/DaoSwap.sol
@@ -18,6 +18,7 @@ contract DaoSwap is Ownable, PullPayment {
     uint8 public swapRandomizationInSecs;
     uint256 public feesAsPct;
     IERC20 token;
+    Stages public stage = Stages.OpenToNewOrders;
 
     // TODO(gp): @toma consider representing the orders as (token1, quantity, token2). In this way
     // We can represent buying 3 wBTC and paying in ETH as (wBTC, 3, ETH), selling 3 wBTC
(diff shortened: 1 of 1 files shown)


defi/dao_swap/contracts/DaoSwap.sol
```diff
@@ -150,6 +168,8 @@ contract DaoSwap is Ownable, PullPayment {
     /// @notice Execute the swap.
     function onSwapTime() public onlyOwner {
         uint256 price;
+        // Change the stage to processing to block new orders
+        stage = Stages.Processing;
```
tamriq, line 172: I just realised that when we change the stage from inside of `onSwapTime()` function, the new stage value is not written to blockchain until `onSwapTime()` is done. So it doesn't make sense :/ My bad. But I have another idea.
E.g. we can have `lockOrders` modifier that locks some functions while the modified function is running.
```
bool private locked;

    modifier lockOrders() {
        locked = true;
        _;
        locked = false;
    }
```
And use it with `onSwapTime`, so the modifier sets locked to true, runs the function, and then sets locked back to false. On the other hand, for sell and buy functions we can create modifier `isLocked`:
```
     modifier isLocked() {
        require(locked == false, "Unable to submit new orders while swap calculations are in progress" )
        _;
    }
```

Makes sense?

tamriq (changes requested):
Good job! But let's change the direction

defi/dao_swap/contracts/DaoSwap.sol
KangmingL: Hi Toma, 
Thanks for your feedback!
I'm a bit confused because from my understanding we want to change the stage value after onSwapTime() is done. Also, what's the difference between using the modifier lockOrders() and changing stages at the beginning and the end of onSwapTime(), I think "_;" just means run the onSwapTime() function.

DanilYachmenev:
probably obsolete, moving to P1 for now

samarth9008:
I believe not required anymore and also the logic of it is changed so closing. If needed we will reopen it.

PR #99, opened 2023-04-01
Issue78 clean up the python reference implementation

DanilYachmenev:
Task #78 

Decided to take out `Order` in a separate file with an intension to use it for both DaoCross and DaoSwap
wdyt?


defi/dao_cross/order.py
```diff
@@ -0,0 +1,160 @@
+"""
+Import as:
+
+import defi.dao_cross.order as ddacrord
+"""
```
DanilYachmenev, line 5: Moved to a separate file with idea that `Order` going to be the same for DaoSwap

defi/dao_cross/test/test_dao_cross.py
```diff
@@ -0,0 +1,110 @@
...
+import pandas as pd
+import helpers.hpandas as hpandas
+import helpers.hunit_test as hunitest
+
+import defi.dao_cross.dao_cross as ddcrdacr
+import defi.dao_cross.order as ddacrord
+
+
+class TestMatchOrders1(hunitest.TestCase):
+    def _get_test_orders(self) -> List[ddacrord.Order]:
```
DanilYachmenev, line 12: We can use `get_random_order()` here with fixed `seed`. Although, function performance will be even less visible

DanilYachmenev:
Since the PR is quite big, let's agree on the code structure here
I'll move necessary classes in smaller PRs first and then do tests in other

gpsaggese (commented):
LG. In general let's separate PR moving stuff / linting from refactoring.

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+
+        :param base_token: token to express order quantity with
+        :param quote_token: token to express order price with
+        :param action: order action type
+            - "buy": purchase the base token and pay with the quote token
+            - "sell": sell the base token and receive the quote token
+        :param quantity: quantity in terms of the base token
+        :param limit_price: limit price in terms of the quote token
+        :param deposit_address: deposit address to implement the order for
+        :param wallet_address: wallet address to implement the order for
```
PomazkinG, line 47: nit: what is the wallet address in terms of the notation from the white-paper?

image

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+        :param timestamp: time of order execution
+        """
+        self.base_token = base_token
+        self.quote_token = quote_token
+        hdbg.dassert_in(action, ["buy", "sell"])
+        self.action = action
+        self.quantity = quantity
+        # Replace NaN with signed `np.inf` depending upon `action`.
+        # This helps with `Order` comparisons (`lt` and `gt`).
+        if np.isnan(limit_price):
```
PomazkinG, line 57: maybe we can use `None` for limit price to indicate that an order is `market order`, i.e. with no limit price

e.g.,
```
limit_price: Optional[float]
```

and here we do:
```
if limit_price is None:
   ...
else: 
   ...
```

not sure how it affects the rest of the code, so even a TODO is fine at this point

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+
+    def __init__(
+        self,
+        base_token: str,
+        quote_token: str,
+        action: str,
+        quantity: float,
+        limit_price: float,
+        deposit_address: Union[int, str],
```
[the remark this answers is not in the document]
PomazkinG: I would even make it `Optional[str]` because most of the time we don't really need an address

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+        wallet_address: Union[int, str],
+        *,
+        timestamp: Optional[datetime.datetime] = None,
+    ) -> None:
+        """
+        Constructor.
+
+        :param base_token: token to express order quantity with
```
gitpaulsmith, line 39: Delete the trailing "with" here and in the next line.

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+        :param quote_token: token to express order price with
+        :param action: order action type
+            - "buy": purchase the base token and pay with the quote token
+            - "sell": sell the base token and receive the quote token
+        :param quantity: quantity in terms of the base token
+        :param limit_price: limit price in terms of the quote token
```
gitpaulsmith, line 45: "in terms of the quote token per base token"

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+        :param deposit_address: deposit address to implement the order for
+        :param wallet_address: wallet address to implement the order for
+        :param timestamp: time of order execution
+        """
+        self.base_token = base_token
```
PomazkinG, line 50: maybe a few fool-proof assertions would help, e.g.,
```
hdbg.dassert_isinstance(base_token, str)
hdbg.dassert_isinstance(quote_token, str)
# Check that quantity is a positive float.
etc
```

defi/dao_cross/order.py
gitpaulsmith: Sounds good to me.

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+        elif self.limit_price > other.limit_price:
+            takes_precedence = True
+        elif self.timestamp > other.timestamp:
+            takes_precedence = True
+        return takes_precedence
+
+
+def get_random_order(seed: Optional[int] = None) -> Order:
+    """
+    Get Order for ETH/BTC with randomized valid parameters.
```
gitpaulsmith, line 128: Use backticks around `Order` since it refers to the class.

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+        action,
+        quantity,
+        limit_price,
+        deposit_address,
+        wallet_address,
+    )
+    return order
+
+
+def is_active_order(order: Optional[Order]) -> bool:
```
gitpaulsmith, line 162: I'd put this in with `dao_cross.py` for now, since it is a helper for implementation.

defi/dao_cross/order.py
```diff
@@ -0,0 +1,172 @@
...
+        self.deposit_address = deposit_address
+        self.wallet_address = wallet_address
+        self.timestamp = timestamp or hdateti.get_current_time(tz="UTC")
+
+    def __repr__(self) -> str:
+        return str(self)
+
+    def __str__(self) -> str:
+        ret = (
+            "base_token=%s quote_token=%s action=%s quantity=%s limit_price=%s timestamp=%s deposit_address=%s"
```
PomazkinG, line 75: nit: I would preserve the order from the ctor

defi/dao_cross_swap/dao_cross.py
```diff
@@ -1,273 +0,0 @@
...
-# %%
-btc_transfers = transfers_df[transfers_df["token"] == "BTC"]
-display(btc_transfers)
-
-# %%
-# Transferred BTC (in BTC).
-btc_transfers["amount"].sum()
-
-# %%
-# BTC transferred divided by the ETH/BTC price (quoted in BTC per ETH)
```
gitpaulsmith, line 253: This reflects an invariant an can (should) be reflected in a unit test. Good that they're preserved in the notebook update (and we can keep them there), but let's also add a test for the invariant.

PomazkinG (commented):
Reviewed the `Order`

I don't have the full context (i.e. how it is used for `DaoSwap`) so you decide what is right

defi/dao_cross/test/test_dao_cross.py
gitpaulsmith: Ok with me to add one or more tests with `get_random_order()`, but since you've already written these explicit tests, I would keep them since they are easy to read.

gitpaulsmith (approved):
LGTM. Feel free to merge and address comments in a follow-up.

defi/dao_cross/order.py
gitpaulsmith: Wallet address is implicit in the white paper. Here we want it explicit for clarity and for the sake of a readable ledger (the output of the reference implementation).

defi/dao_cross/order.py
gitpaulsmith: Here we can keep `limit_price` in the realm of float. IIRC an earlier version of a draft of `Order` used `np.nan`. Not an essential detail.

defi/dao_cross/order.py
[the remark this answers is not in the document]
gitpaulsmith: We do need an address for the output ledger. To keep things simple, the random order generates `int` addresses randomly. Of course, in the real implementation, the address will have a strict form.

defi/dao_cross/order.py
gitpaulsmith: Agree.

defi/dao_cross/order.py
gitpaulsmith: I think this has been changing. I would (in a follow-up) move to the constructor ordering in the white paper, with the source wallet address tacked on at the end.

defi/dao_cross/order.py
DanilYachmenev: decided to get rid of such cases actually and refer to an order just as a word directly
so we do not need to put to many backticks, since the point is conveyed just by "order" word

PR #100, opened 2023-04-03
Intraday Volume Predicition task2

dchoi127:
Implemented OLS model for volume.


gpsaggese (commented):
Sorry for the delay. We usually respond to PRs in 24 hrs.
Web3 projects got funded so been super busy with that.
 
Good job. Let's discuss the results in the next office hours

PR #101, opened 2023-04-03
Sorr task80 move code to lib

PomazkinG:
Freezing the behavior before refactoring. The next step is to extend for N orders.

@gpsaggese @gitpaulsmith FYI

@DanilYachmenev FYI: copy-pasted the `Order` class from your PR https://github.com/sorrentum/sorrentum/pull/99 to keep moving


gpsaggese (commented):
LG

Issue #82, opened 2023-03-29
Implement Uniswap reference model in Python


tamriq:
Compute impermanent loss

gpsaggese:
Correct.

We want to implement a Python code that "simulates" the behavior of UniswapV1
A decent explanation is https://jeiwan.net/posts/programming-defi-uniswap-1/ and related blog posts.

PR1: Implement UniswapV1
```
class UniswapV1

    def addLiquidity()

    def getPrice()

    ...
```

We don't want to "simulate" the swap (i.e., transferring tokens) but only getting prices and updating the state of the contract. Makes sense?

We want to create unit tests to make sure things are fine.

PR2: Then we want to create a special unit tests that "simulates" people adding / removing liquidity returning the prices.

Let's do small and quick iterations as always.

DanilYachmenev:
@gpsaggese I've filed a couple of tasks with unit tests for students to write - should be a good practice for them
So I've delegated unit tests for Uniswap as well here
https://github.com/sorrentum/sorrentum/issues/111

I think nothing difficult there, I'll help guys if needed
So if this is OK, I think I can start to work on the next steps with Uniswap. like writing conversion function or to compute impermanent loss

wdyt?

gpsaggese:
Yes, all unit tests can be outsourced as first bugs. 
For each "class" of unit tests can you add one example of unit tests and then people can copy-paste-refactor and change it to add more? I've marked those tasks as `Dan -> Outsource`

Can you pls strike out the filed bug on https://docs.google.com/document/d/1Kxu6gfUTtfLKaw2O_SBn_iBfpORhlRFMB4uJ4RpA4Hs/edit#?

As to next steps, this paper https://arxiv.org/abs/2208.06046 (a bit technical) gave us a lot of good ideas on how to compare our stuff to Uniswap. We will have a clear plan in the next few days.
Feel free to coordinate with @tamriq  and @PomazkinG what other bug you can attack from the list.

PR #103, opened 2023-04-04
Issue82: Add Uniswap class

DanilYachmenev:
Issue #82 

Implemented with reference to https://jeiwan.net/posts/programming-defi-uniswap-1/ 

Note that I've created `get_token_amount()` and `get_eth_amount()` instead of `getPrice()` since from what I've read Uniswap is primarily a token exchange platform, so it generally deals in the amount of tokens being exchanged rather than prices - 1 function would suit only for 1 token

### DIFF
--- a/defi/uni_swap/uni_swap.py
+++ b/defi/uni_swap/uni_swap.py
@@ -1,0 +1,73 @@
+"""
+Import as:
+
+import defi.uni_swap.uni_swap as duswunsw
+"""
+
+import helpers.hdbg as hdbg
+
+
+class UniswapV1:
+    """
+    Simulate behavior of Uniswap.
+
+    Calculations are based on the Uniswap formula of constant product,
+    i.e. if one token's balance in the pool changes, the other token's
+    balance must change to the opposite direction by a proportional
+    amount to maintain the constant product.
+    """
+
+    def __init__(self, token_reserve: float, eth_reserve: float):
+        """
+        Constructor.
+
+        :param token_reserve: token reserve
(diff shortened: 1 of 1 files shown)


defi/uni_swap/uni_swap.py
```diff
@@ -0,0 +1,73 @@
...
+
+        :param token_reserve: token reserve
+        :param eth_reserve: ETH reserve
+        """
+        self.token_reserve = token_reserve
+        self.eth_reserve = eth_reserve
+
+    def add_liquidity(self, token_amount: float, eth_amount: float) -> None:
+        """
+        Adds liquidity by increasing the token and ETH reserves.
```
gpsaggese, line 32: -> Add

defi/uni_swap/uni_swap.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def _get_amount(
+        self, input_amount: float, input_reserve: float, output_reserve: float
+    ) -> float:
+        """
+        Get the amount of output token.
+
+        :param input_amount: input token reserve
+        :param input_reserve: output token reserve
+        :param output_reserve: output token reserve
```
gpsaggese, line 65: Can you fix / improve this?

gpsaggese (commented):
Minor changes

defi/uni_swap/uni_swap.py
DanilYachmenev: extended a bit, do not want to mix with class description
do you want to describe process in more detail?

tamriq (approved):

Issue #80, opened 2023-03-29
Implement DaoSwap optimization problem


tamriq:
From whitepaper:
>The general problem of determining the token equilibrium price and the allocation among the swap participants can be formulated in terms of the inequalities on quantities and prices corresponding to the orders and on the need that the total quantity exchanged of each token is preserved across the swaps, i.e., the quantity bought for each token is equal to the sold quantity.

> The general solution of the DaoSwap problem is a set of non-linear equation with combinatorial constraints, which is of course NP-complete.
It can still be solved in an effective way with solvers like ...

- Modeling languages
https://coin-or.github.io/pulp/
https://pyomo.readthedocs.io/en/latest/index.html
https://stackoverflow.com/questions/26305704/python-mixed-integer-linear-programming
Questions:
Do they support open-source backends?

- Use one of the solvers (open-source)
https://www.gnu.org/software/glpk/
https://github.com/coin-or/Cbc
https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.milp.html
http://users.isy.liu.se/johanl/yalmip/
Closed-source: CPLEX and Mosek (used in the past, but too expensive)

- How to represent non-linear constraints
https://math.stackexchange.com/questions/2117912/conditional-constraint-in-a-linear-program
https://stackoverflow.com/questions/57949190/how-to-apply-conditional-constraints-to-python-pulp-function
Lots of examples of how to map non-linear constraints onto CPLEX (probably possible to reformulate for other solvers)
https://www.linkedin.com/pulse/making-optimization-simple-python-alex-fleischer/

- Interesting papers
https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=c66efc96ff4099fa214dfef684cdddd30b0fc770

gpsaggese:
I can work on this with @PomazkinG when he has bandwidth

PomazkinG:
@gpsaggese I could use some help here

What I don't quite get is what do we want to maximize / minimize?

Typically the `pulp` solver accepts:
- The objective function (i.e. what we want to maximize / minimize)
   - e.g., z = f(x, y, z)
- Constraints on x, y, z
   - e.g., 10x + 20y = 50 and 5x - 2z <= 20  

Now let's come back to our toy example:
- 2 limit orders
- opposite direction
- same base token = "BTC"
- prices are known

[...]

PR #105, opened 2023-04-04
Sorr task80 extend for n orders

PomazkinG:
#80 

Extending the function to work with N orders


PomazkinG:
@gitpaulsmith @gpsaggese FYI

PR #106, opened 2023-04-05
Sorr task80 small improvements

PomazkinG:
- Apparently the big M method just created unnecessary complexity -> converted into the Python-native `if-else`
- Updated the tests to highlight the fact that if the limit price condition is False for an order then the exchanged quantity is zero only for this particular order, not for all


gpsaggese (commented):
LG

PR #124, opened 2023-04-12
Issue10 create a docker script to clean up/format solidity

samarth9008:


dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,88 @@
...
+#
+# Prettier formatter
+# sudo npm install --save-dev prettier prettier-plugin-solidity
+# npx prettier --write \'contracts/**/*.sol\'
+#
+# EthLint
+# sudo npm install -g solium
+# solium --init
+# solium -c --dir contracts . 
+# Usage: ./lint_solidity.sh file1.sol file2.sol ....
```
gpsaggese, line 13: Let's put the usage right after line 3.

Linux command are like
```
# Usage:
# > lint_solidity.sh ...
```

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env bash
+#
+# Dockerized solidity formatter and linter
+#
+# Prettier formatter
+# sudo npm install --save-dev prettier prettier-plugin-solidity
```
gpsaggese, line 6: Need > but I would argue that the info about running prettier and solium are implementation details and should not be in the interface documentation

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,88 @@
...
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Install Node.js
+RUN curl -sL https://deb.nodesource.com/setup_18.x | bash -
+RUN apt-get update && apt-get install -y nodejs
+RUN node -v
+RUN npm -v
+
+#Install Prettier and Ethlint(Solium)
```
gpsaggese, line 33: Comments need a space and end with .

```
# Install Prettier and Ethlint / Solium.
```

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,88 @@
...
+#
+# EthLint
+# sudo npm install -g solium
+# solium --init
+# solium -c --dir contracts . 
+# Usage: ./lint_solidity.sh file1.sol file2.sol ....
+
+set -eux
+
+cat >/tmp/tmp.dockerfile <<EOF
```
gpsaggese, line 17: This file should be unique to the script otherwise there might be a collision with other people running similar scripts.

TMP_FILENAME="/tmp/tmp.lint_solidity.Dockerfile"

Also tmp files should have info on who generates it

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env bash
+#
+# Dockerized solidity formatter and linter
```
gpsaggese, line 3:
```
# Solidity formatter and linter using prettier and EthLint.
```

IMO no need to say Dockerized since it's an implementation detail

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,88 @@
...
+EOF
+
+docker build -f /tmp/tmp.dockerfile -t ubuntulint .
+
+USER="$(id -u $(logname)):$(id -g $(logname))"
+WORKDIR="$(realpath .)"
+MOUNT="type=bind,source=${WORKDIR},target=${WORKDIR}"
+
+: <<'END_COMMENT'
+Since solium does not support multiple file inputs but a directory
```
gpsaggese, line 50: Let's use normal comments

```
# Since Ethlint doesn't does not support multiple file inputs but only directories,
# we copy all the files supplied to the command line to a temporary directory and
# use it as input to Ethlint.
```

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,88 @@
...
+WORKDIR="$(realpath .)"
+MOUNT="type=bind,source=${WORKDIR},target=${WORKDIR}"
+
+: <<'END_COMMENT'
+Since solium does not support multiple file inputs but a directory 
+so copied all the files supplied to command line arguments  
+to temporary created contracts directory which will work as input to ethlint
+END_COMMENT
+
+mkdir temp
```
gpsaggese, line 55: -> 
TMP_DIR_NAME="./tmp.lint_solidity"

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,88 @@
...
+# Loop through each file name passed as an argument
```
gpsaggese, line 56: argument.

gpsaggese (commented):
Very good job. Mainly readability comments

Let's do another iteration and then it's merge

dev_scripts/lint_solidity.sh
samarth9008: I am not sure on how to solve this `Also tmp files should have info on who generates it`

gpsaggese: Just adding info in the name of which script generated is enough (like tmp.lint_solidity.Dockerfile instead of tmp.Dockerfile)

samarth9008:
Made the required changes.

gpsaggese:
Sg. The protocol is to push the "Review button again".

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,83 @@
...
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Install Node.js.
+RUN curl -sL https://deb.nodesource.com/setup_18.x | bash -
+RUN apt-get update && apt-get install -y nodejs
+RUN node -v
+RUN npm -v
+
+#Install Prettier and Ethlint/Solium.
```
gpsaggese, line 27: You forgot the space before Install here

dev_scripts/lint_solidity.sh
```diff
@@ -0,0 +1,83 @@
...
+
+EOF
+
+docker build -f $TMP_FILENAME -t ubuntulint .
+
+USER="$(id -u $(logname)):$(id -g $(logname))"
+WORKDIR="$(realpath .)"
+MOUNT="type=bind,source=${WORKDIR},target=${WORKDIR}"
+
+# Since Ethlint doesn't does not support multiple file inputs but only directories,
```
gpsaggese, line 43: doesn't does not -> doesn't

gpsaggese (commented):
One small one iteration

gpsaggese (commented):
Good work

PR #125, opened 2023-04-12
Task79: implement DaoCross 0.1

tamriq:
- heap data structure are moved to the separate library, since this code is going to be reused in DaoSwap
- mock contracts for ERC20 and price oracle for easier testing
- small tests for matchOrders function - to check output transfers


gpsaggese (commented):
LG. I'll read line by line today

Issue #139, opened 2023-04-18
Create some pathological cases for supply / demand curve


DanilYachmenev:
Create functions to generate orders that give a certain demand / supply curve shape and then one can call it to create many different problems.

We need all cases kike
- [x] Multiple intersection points at quantity Q'
- [x] Multiple intersection points at price P'
- [x] No intersection

Note: there are no cases for these type of curves on which there will be multiple intersection points at multiple Q and P simultaneously. At intersection point (p0, q0) the next demand P and Q are always not higher than (p0, q0) while supply P and Q are always not less than (p0, q0)

`get_supply_curve1() -> List[Order]:`
E.g.
<img width="487" alt="Снимок экрана 2023-04-18 в 06 57 07" src="https://user-images.githubusercontent.com/31514660/232635209-474a5ea0-4d14-465a-9e12-277469a9de27.png">
<img width="444" alt="Снимок экрана 2023-04-18 в 06 57 19" src="https://user-images.githubusercontent.com/31514660/232635227-b2419d5e-5ea0-4f89-84c7-b4f15bf36e91.png">

FYI @tamriq

DanilYachmenev:
Done, closing

PR #153, opened 2023-04-19
Issue139: Create supply demand curves

DanilYachmenev:
Task #139 

DRAFT
Will show demonstration of how it works in a notebook

### DIFF
--- a/defi/dao_cross/supply_demand.py
+++ b/defi/dao_cross/supply_demand.py
@@ -1,0 +1,169 @@
+"""
+Import as:
+
+import defi.dao_cross.supply_demand as ddcrsede
+"""
+
+import copy
+import heapq
+import logging
+from typing import Any, Dict, List, Optional
+
+import pandas as pd
+
+import defi.dao_cross.order as ddacrord
+import helpers.hdbg as hdbg
+import helpers.hprint as hprint
+
+_LOG = logging.getLogger(__name__)
+
+
+def get_supply_orders1() -> List[ddacrord.Order]:
+    ts = None
+    action = "sell"
+    base = "BTC"
+    quote = "ETH"
+    src = -1
+    dst = -1
+    orders = [
+        ddacrord.Order(
+            ts, action, quantity=40, base, limit_price=100, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=40, base, limit_price=60, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=30, base, limit_price=40, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=30, base, limit_price=30, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=20, base, limit_price=20, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=20, base, limit_price=10, quote, src, dst
+        ),
+    ]
+    return orders
+
+
+def get_supply_orders2() -> List[ddacrord.Order]:
+    ts = None
+    action = "sell"
+    base = "BTC"
+    quote = "ETH"
+    src = -1
+    dst = -1
+    orders = [
+        ddacrord.Order(
+            ts, action, quantity=4, base, limit_price=10, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=4, base, limit_price=6, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=3, base, limit_price=4, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=3, base, limit_price=3, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=2, base, limit_price=2, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=2, base, limit_price=1, quote, src, dst
+        ),
+    ]
+    return orders
+
+
+def get_demand_orders1() -> List[ddacrord.Order]:
+    ts = None
+    action = "buy"
+    base = "BTC"
+    quote = "ETH"
+    src = 1
+    dst = 1
+    orders = [
+        ddacrord.Order(
+            ts, action, quantity=10, base, limit_price=110, quote, src, dst
+        ),
+        ddacrord.Order(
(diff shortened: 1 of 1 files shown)


defi/dao_cross/supply_demand.py
```diff
@@ -0,0 +1,169 @@
...
+            ts, action, quantity=20, base, limit_price=20, quote, src, dst
+        ),
+        ddacrord.Order(
+            ts, action, quantity=20, base, limit_price=10, quote, src, dst
+        ),
+    ]
+    return orders
+
+
+def get_supply_orders2() -> List[ddacrord.Order]:
```
gpsaggese, line 51: This seems the same as the one above.

How about passing a param like quantity_scale, quantity_const and limit_price_scale so that one can pass a value that is used to "move" the curve around

and then

quantity=4 * quantity_scale + quantity_const

In this way you don't have to implement one function per curve but it is parametrized

defi/dao_cross/supply_demand.py
```diff
@@ -0,0 +1,169 @@
...
+
+import defi.dao_cross.order as ddacrord
+import helpers.hdbg as hdbg
+import helpers.hprint as hprint
+
+_LOG = logging.getLogger(__name__)
+
+
+def get_supply_orders1() -> List[ddacrord.Order]:
+    ts = None
```
gpsaggese, line 22: A short description like:
```
Get orders corresponding to a monotonically decreasing supply curve.
```

defi/dao_cross/supply_demand.py
```diff
@@ -0,0 +1,169 @@
...
+            ts, action, quantity=3, base, limit_price=1, quote, src, dst
+        ),
+    ]
+    return orders
+
+
+def get_curve(
+    orders: List[ddacrord.Order], type_: str
+) -> List[Tuple[int, int]]:
+    # Extract quantity and proce from the passed orders in order to filter them.
```
gpsaggese, line 144: -> price

defi/dao_cross/supply_demand.py
```diff
@@ -0,0 +1,169 @@
...
+    orders_info = [
+        (order.quantity, order.limit_price,) for order in orders
+    ]
+    # Sort orders by limit price so order of asks is preserved.
+    if type_ == "supply":
```
gpsaggese, line 149: ascending = type_ == "supply"
orders_info = sorted(..., ascending=ascending)

defi/dao_cross/supply_demand.py
```diff
@@ -0,0 +1,169 @@
...
+        # Supply curve starts from the order with the lowest limit price.
+        orders_info = sorted(orders_info, key=lambda x: x[1])
+    elif type_ == "demand":
+        # Supply curve starts from the order with the highest limit price.
+        orders_info = sorted(orders_info, key=lambda x: x[1], ascending=False)
+    else:
+        ValueError("Invalid type_='%s'" % self.type_)
+    # Get the coordinates where curves crosses prices axis.
+    first_curve_point = (0, orders_info[0][1],)
```
gpsaggese, line 158: No need for the last ,

defi/dao_cross/supply_demand.py
```diff
@@ -0,0 +1,169 @@
...
+    # Initiate the list with the first coordintate.
+    curve_points = [first_curve_point,]
```
gpsaggese, line 160: No need for ,

gpsaggese (commented):
Going in the right direction. Some suggestion

Agree that having a notebook to show the curves will be helpful

defi/dao_cross/supply_demand.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def _get_curve_orders(
+    action: str,
+    quantities: List[int],
+    limit_prices: List[int],
+    *,
+    quantity_scale: float = 1.0,
+    quantity_const: float = 0.0,
+    limit_price_scale: float = 1.0,
```
DanilYachmenev, line 28: maybe we would like to use only constants to just move the lines?
Their usage is clear, while scale-type params may have limited use

So maybe just these instead?
```
quantity_const: float  = 0.0,
limit_price_const: float  = 0.0,
```

defi/dao_cross/notebooks/SorrTask139_Create_some_pathological_cases_for_supply_demand_curve.py
```diff
@@ -0,0 +1,100 @@
...
+
+# %%
+type_ = "demand"
+demand_curve = ddcrsede.get_curve(demand_orders, type_)
+demand_curve
+
+# %%
+plt.plot(*zip(*supply_curve))
+plt.plot(*zip(*demand_curve))
+plt.show()
```
DanilYachmenev, line 68: And actually there is a way to find all the EXACT intersection points between 2 lines like that: [Bentley Ottmann Algorithm](https://www.hackerearth.com/practice/math/geometry/line-intersection-using-bentley-ottmann-algorithm/tutorial/#:~:text=Line%20intersection%20problem%20is%20a,intersections%20between%20these%20line%20segments.)

I played around with it a bit, we can do it
but do we need it?

Or we'd better not focus on it and better work with [linear representations of the curves?](https://github.com/sorrentum/sorrentum/issues/154)

But we probably want to have a way to detect the pathological cases algorithmically, not only visually, right?

DanilYachmenev:
Actually, found some inaccuracies during integration of optimizer code.
Will fix overnight and simplify the examples to make the checks visible

defi/dao_cross/supply_demand.py
```diff
@@ -0,0 +1,106 @@
...
+            limit_price,
+            quote_token,
+            deposit_address,
+            wallet_address,
+        )
+        orders.append(order)
+    return orders
+
+
+def get_curve_dots(
```
gitpaulsmith, line 64: Suggest we return a pandas series (and we can use `type_` as the name) to make manipulations slightly easier (we have a pandas dependency in `Order` anyway (as a timestamp)).

defi/dao_cross/supply_demand.py
gitpaulsmith: I'd just keep it as-is for now since you've implemented that flexibility.

gitpaulsmith (approved):

DanilYachmenev:
Merging since cases are agreed
Further changes will be implemented in a follow-up

Issue #168, opened 2023-04-25
Allow notebook publishing at Sorrentum


DanilYachmenev:
Currently if we try to publish a notebook from `sorrentum`:
```
While executing 'get_html_dir_to_url_mapping()' caught error:
name 'get_html_dir_to_url_mapping' is not defined
Trying to continue
Traceback (most recent call last):
  File "/app/dev_scripts/notebooks/publish_notebook.py", line 351, in <module>
    _main(_parse())
  File "/app/dev_scripts/notebooks/publish_notebook.py", line 333, in _main
    url_bucket_path = dir_to_url[html_bucket_path]
TypeError: 'NoneType' object is not subscriptable
```

Fix is just to add `get_html_dir_to_url_mapping()` to `repo_config.py`

PR #169, opened 2023-04-25
SorrTask168: Allow notebook publishing at sorrentum

DanilYachmenev:
#168 

This allows us to publish notebooks from sorrentum now
used at https://github.com/sorrentum/sorrentum/pull/153#issuecomment-1521324770


tamriq (approved):

repo_config.py
```diff
@@ -357,6 +357,18 @@ def get_html_bucket_path() -> str:
     return html_bucket_path
 
 
+def get_html_dir_to_url_mapping() -> Dict[str, str]:
```
tamriq, line 360: should't we add the name of the new function to `config_func_to_str()`?

gpsaggese, line 360: LG for now.

We should have a way to publish the notebook in a Sorrentum bucket so that we can share notebooks with people without access to our VPN. Maybe we can have a single API key for all the Sorrentum contributors or (more painfully to set up) one key per contributors

@jsmerix feel free to file a P1 bug on this

PR #191, opened 2023-05-01
SorrTask148: clean and lint DaoCross

tamriq:
1) matchOrders function was renamed to processOrders and separated to 3 functions
    - filterOrders
    - matchOrders
    - returnLeftovers
2) all contracts in dao_cross were linted


defi/dao_cross/contracts/PriceOracle.sol
```diff
@@ -23,12 +21,13 @@ contract PriceOracle {
...
-            /*uint updatedAt*/,
-            /*uint80 answeredInRound*/
-        ) = priceFeed.latestRoundData();
+            ,
+            /* uint80 roundID */ int256 price,
+            ,
+            ,
+
+        ) = /*uint startedAt*/ /*uint updatedAt*/ /*uint80 answeredInRound*/
+            priceFeed.latestRoundData();
```
gpsaggese, line 30: This is a bit weird, do you know if `//` would be better for the linter?

PR #266, opened 2023-05-18
SorrTask260: indexer to catch tulip order events

tamriq:
1) reorg twap adapter and indexer to the separate tulip/services dir
2) indexer script + Dockerfile
3) indexer catches the buy/sell orders events from the tulip contract and puts them to the database


tamriq:
@jsmerix Could please have a quick look on the script and dockerfile when u have time? The code is pretty simple, it just monitors blockchain for specific events, but it's very important for us to make it stable and not to loose anything. Maybe u are familiar for best practices for providing the stability, would be useful to hear some advices from u

jsmerix (approved):
@tamriq I have added a couple of suggestions which might be worth taking a look at.

Issue #160, opened 2023-04-21
Create a Python script to convert a Parquet file to CSV


PomazkinG:
From @gpsaggese 

> To bridge the gap of observability between Parquet and humans, it might be useful to have a simple script parquet_to_csv.py that converts a Parquet file in a CSV file or stdin for quick inspection and piping to https://csvkit.readthedocs.io/en/latest/
Can you pls file a bug on the Sorrentum side and put a label "good-first-bug" so we outsource it to students

FYI: there are some scripts that we can re-use (?), e.g.,:
- https://github.com/sorrentum/sorrentum/blob/master/im_v2/common/data/transform/convert_csv_to_pq.py
- https://github.com/sorrentum/sorrentum/blob/master/im_v2/common/data/transform/pq_convert.py

Add more specs here

PomazkinG:
> Both links seem to be pointing to private or non existing repos. Are there any other sources you could provide?

Updated the original post, should be public now

dchoi127:
[PR](https://github.com/sorrentum/sorrentum/pull/267)

gpsaggese:
I've just noticed this. @dchoi127 you've assigned the review to yourself so we didn't get notified.

Routing to @DanilYachmenev for taking the first look (@PomazkinG feel free to also review)

DanilYachmenev:
So guys lets talk about specs a bit

1. Following, https://github.com/sorrentum/sorrentum/pull/267#discussion_r1201092684, did I get it right that we want to build a script that converts 1 Parquet file to 1 CSV file?
2. If yes, then I have a question about `aws_profile`

Basically we have 2 files (input and output) and 1 `aws_profile` now

If we want to implement conversion just locally or only inside AWS, then everything is easy
But practically we may want to convert a `.pq` file that is stored locally to a location at AWS or between different AWS profiles

Should we care about that?
If yes, we can pass 2 AWS profiles: `src_aws_profile` and `dst_aws_profile`
Both can be optional (default is `None` which would mean that file is stored locally)
IMO it is going to look OK as well and not complicating interface too much

@PomazkinG, @gpsaggese wdyt?

PomazkinG:
my 2 cents: I would start with implementing a basic functionality and extend it depending on use cases.

E.g., we start with 1 AWS profile but when we face an issue where we need to use the 2nd AWS profile we file a separate issue and extend the functionality.

gpsaggese:
[...]

PR #267, opened 2023-05-19
Parquet to CSV

dchoi127:
#160 

Converts all parquet files in the given target dir 

Followed similar approach from https://github.com/sorrentum/sorrentum/blob/master/im_v2/common/data/transform/convert_csv_to_pq.py

### DIFF
--- a/devops/compose/docker-compose.yml
+++ b/devops/compose/docker-compose.yml
@@ -1,0 +1,98 @@
+version: '3'
+
+services:
+  base_app:
+    cap_add:
+      - SYS_ADMIN
+    environment:
+      - AM_AWS_ACCESS_KEY_ID=$AM_AWS_ACCESS_KEY_ID
+      - AM_AWS_DEFAULT_REGION=$AM_AWS_DEFAULT_REGION
+      - AM_AWS_PROFILE=$AM_AWS_PROFILE
+      - AM_AWS_S3_BUCKET=$AM_AWS_S3_BUCKET
+      - AM_AWS_SECRET_ACCESS_KEY=$AM_AWS_SECRET_ACCESS_KEY
+      - AM_ECR_BASE_PATH=$AM_ECR_BASE_PATH
+      - AM_ENABLE_DIND=0
+      - AM_FORCE_TEST_FAIL=$AM_FORCE_TEST_FAIL
+      - AM_HOST_NAME=Davids-MacBook-Pro-2.local
+      - AM_HOST_OS_NAME=Darwin
+      - AM_HOST_USER_NAME=davidchoi
+      - AM_HOST_VERSION=22.1.0
+      - AM_REPO_CONFIG_CHECK=True
+      # Use inferred path for `repo_config.py`.
+      - AM_REPO_CONFIG_PATH=
+      - AM_TELEGRAM_TOKEN=$AM_TELEGRAM_TOKEN
+      - CK_AWS_ACCESS_KEY_ID=$CK_AWS_ACCESS_KEY_ID
+      - CK_AWS_DEFAULT_REGION=$CK_AWS_DEFAULT_REGION
+      - CK_AWS_PROFILE=$CK_AWS_PROFILE
+      - CK_AWS_S3_BUCKET=$CK_AWS_S3_BUCKET
+      - CK_AWS_SECRET_ACCESS_KEY=$CK_AWS_SECRET_ACCESS_KEY
+      - CK_ECR_BASE_PATH=$CK_ECR_BASE_PATH
+      # - CK_ENABLE_DIND=
+      # - CK_FORCE_TEST_FAIL=$CK_FORCE_TEST_FAIL
+      # - CK_HOST_NAME=
+      # - CK_HOST_OS_NAME=
+      # - CK_PUBLISH_NOTEBOOK_LOCAL_PATH=$CK_PUBLISH_NOTEBOOK_LOCAL_PATH
+      - CK_TELEGRAM_TOKEN=$CK_TELEGRAM_TOKEN
+      - GH_ACTION_ACCESS_TOKEN=$GH_ACTION_ACCESS_TOKEN
+      # This env var is used by GH Action to signal that we are inside the CI.
+      - CI=$CI
+    image: ${IMAGE}
+
+    restart: "no"
+    volumes:
+      # TODO(gp): We should pass the value of $HOME from dev.Dockerfile to here.
+      # E.g., we might define $HOME in the env file.
+      - ~/.aws:/home/.aws
+      - ~/.config/gspread_pandas/:/home/.config/gspread_pandas/
+      - ~/.config/gh:/home/.config/gh
+
+      # Use sibling-container approach.
+      - /var/run/docker.sock:/var/run/docker.sock
+
+  # Mount `amp` when it is used as supermodule.
+  app:
+    extends:
+      base_app
+    volumes:
+      - ../../:/app
+
+  linter:
+    extends:
+      base_app
+    volumes:
+      - /Users/davidchoi/src/sorrentum1:/src
+      - ../../:/app
+    working_dir: /src
+    environment:
+      - MYPYPATH
+
+  jupyter_server:
+    command: devops/docker_run/run_jupyter_server.sh
+    environment:
+      - PORT=${PORT}
+    extends:
+      app
+    network_mode: ${NETWORK_MODE:-bridge}
+    ports:
+      # TODO(gp): Rename `AM_PORT`.
+      - "${PORT}:${PORT}"
+
+  # TODO(gp): For some reason the following doesn't work.
+  #  jupyter_server_test:
+  #    command: jupyter notebook -h 2>&1 >/dev/null
+  #    extends:
+  #      jupyter_server
+
+  jupyter_server_test:
+    command: jupyter notebook -h 2>&1 >/dev/null
+    environment:
+      - PORT=${PORT}
+    extends:
+      app
+    network_mode: ${NETWORK_MODE:-bridge}
+    ports:
+      - "${PORT}:${PORT}"
+
+networks:
+  default:
+    name: main_network
--- a/im_v2/common/data/transform/parquet_to_csv.py
+++ b/im_v2/common/data/transform/parquet_to_csv.py
@@ -1,0 +1,91 @@
+import argparse
+import logging
+import os
+
+import pandas as pd
+
+
+import helpers.hdbg as hdbg
+import helpers.hparquet as hparque
+import helpers.hparser as hparser
+import helpers.hs3 as hs3
+import im_v2.common.data.transform.transform_utils as imvcdttrut
+
+_LOG = logging.getLogger(__name__)
+
+def _run(args: argparse.Namespace) -> None:
+    if args.aws_profile is not None:
+        hs3.get_s3fs(args.aws_profile)
+    else:
+        hparser.create_incremental_dir(args.dst_dir, args)
+    files = hs3.listdir(
+        args.src_dir,
+        "*.pq*",
+        only_files=True,
+        use_relative_paths=True,
+        aws_profile=args.aws_profile,
+    )
+    _LOG.info("Files found at %s:\n%s", args.src_dir, "\n".join(files))
+    for file in files:
+        full_path = os.path.join(args.src_dir, file)
+        _LOG.debug("Converting %s...", full_path)
+
+
+        df = pd.read_parquet(full_path)
+        df.to_csv(full_path)
+
+        # # Set datetime index.
+        # reindexed_df = imvcdttrut.reindex_on_datetime(df, args.datetime_col)
+        # # Add date partition columns to the dataframe.
+        # df, partition_cols = hparque.add_date_partition_columns(
+        #     reindexed_df, "by_year_month"
+        # )
+        # hparque.to_partitioned_parquet(
+        #     df,
+        #     [args.asset_col] + partition_cols,
+        #     args.dst_dir,
+        #     aws_profile=args.aws_profile,
+        # )
+    #hparque.list_and_merge_pq_files(args.dst_dir, aws_profile="ck")
+
+
+def _parse() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+    parser.add_argument(
+        "--src_dir",
+        action="store",
+        type=str,
+        required=True,
+        help="Dir with input Parquet files to convert to CSV format",
+    )
+    parser.add_argument(
+        "--dst_dir",
+        action="store",
+        type=str,
+        required=True,
+        help="Destination dir where to save converted CSV files",
+    )
+    parser.add_argument(
+        "--incremental",
+        action="store_true",
+        help="Skip files that have already been converted",
+    )
+    parser.add_argument(
+        "--aws_profile",
+        action="store",
+        required=True,
+        type=str,
+        help="The AWS profile to use for .aws/credentials or for env vars",
+    )
+    parser = hparser.add_verbosity_arg(parser)
+    return parser
+
+def _main(parser: argparse.ArgumentParser) -> None:
+    args = parser.parse_args()
+    hdbg.init_logger(verbosity=args.log_level, use_exec_path=True)
+    _run(args)
+
+if __name__ == "__main__":
+    _main(_parse())


gpsaggese (commented):
Going in the right direction but first PR is always painful

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
+import argparse
```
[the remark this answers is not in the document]
dchoi127: I'm not sure how to run that.

https://rikwatson.github.io/python_lint

is this what we're looking for?

devops/compose/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
+version: '3'
+
+services:
+  base_app:
```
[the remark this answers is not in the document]
dchoi127: changed PR name

im_v2/common/data/transform/parquet_to_csv.py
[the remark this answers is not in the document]
DanilYachmenev: pls read [First_review_process](https://github.com/sorrentum/sorrentum/blob/master/docs/First_review_process.md) doc, there you can find a detailed info on how to run linter and some more useful advices

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
...
+    parser.add_argument(
+        "--dst_dir",
+        action="store",
+        type=str,
+        required=True,
+        help="Destination dir where to save converted CSV files",
+    )
+    parser.add_argument(
+        "--incremental",
+        action="store_true",
```
DanilYachmenev, line 72: lets add a line with 
`required=True,`

and put it at the end
in general, place optional params and vars after the required ones

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
...
+        # df, partition_cols = hparque.add_date_partition_columns(
+        #     reindexed_df, "by_year_month"
+        # )
+        # hparque.to_partitioned_parquet(
+        #     df,
+        #     [args.asset_col] + partition_cols,
+        #     args.dst_dir,
+        #     aws_profile=args.aws_profile,
+        # )
+    #hparque.list_and_merge_pq_files(args.dst_dir, aws_profile="ck")
```
DanilYachmenev, line 49: what about this part?

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
...
+        only_files=True,
+        use_relative_paths=True,
+        aws_profile=args.aws_profile,
+    )
+    _LOG.info("Files found at %s:\n%s", args.src_dir, "\n".join(files))
+    for file in files:
+        full_path = os.path.join(args.src_dir, file)
+        _LOG.debug("Converting %s...", full_path)
+
+
```
DanilYachmenev, line 33: do not leave empty lines inside function bodies

im_v2/common/data/transform/parquet_to_csv.py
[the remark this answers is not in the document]
PomazkinG: @samarth9008 no need to update then, as Dan mentioned there is already a section in the docs, so all good

https://github.com/sorrentum/sorrentum/blob/master/docs/First_review_process.md

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
...
+        df = pd.read_parquet(full_path)
+        df.to_csv(full_path)
```
DanilYachmenev, line 35: well, here you read a `.pq` filename and then you write `.csv` file to exactly the same location - ain't gonna work

you can use parquet filename and convert it to csv like
`file.pq` -> `file.csv`

then `csv_path = os.path.join(src_dir, csv_filename)` 
and then `df.to_csv(csv_path)`

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
...
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+    parser.add_argument(
+        "--src_dir",
+        action="store",
+        type=str,
+        required=True,
+        help="Dir with input Parquet files to convert to CSV format",
+    )
```
DanilYachmenev, line 62: Agree with GP comment here https://github.com/sorrentum/sorrentum/pull/267/files#r1201094843

So here lets do `pq_path` here instead of `src_dir` and then just convert 1 specified file

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
...
+    parser.add_argument(
+        "--dst_dir",
+        action="store",
+        type=str,
+        required=True,
```
DanilYachmenev, line 67: lets make it optional, and `None` by default

if no value is specified, save at the same location as `pq_path` is placed
just remember bout the name

in most cases we'd just be able to call a script with specifying only a file name and then it is stored tight next to the passed file

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
+import argparse
+import logging
+import os
+
```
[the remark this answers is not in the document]
DanilYachmenev: and let's move this script from `im_v2/common/data/transform/` to `dev_scripts` since it is quite general
@PomazkinG wdyt?

DanilYachmenev (commented):
good first step, I see that you've read the ref files

see the comments above and below for further improvements

And most importantly, pls try to run this script yourself from the cmd line and pput the logs here so we could see that it works

devops/compose/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
+version: '3'
```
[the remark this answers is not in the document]
dchoi127: Do you mean that I don't want to add the docker-compose.yml to the pr?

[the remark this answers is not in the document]
gpsaggese: Correct. That file is generated automatically so it should not be checked in

In general one should only add files that you have created / you know are needed

im_v2/common/data/transform/parquet_to_csv.py
dchoi127: This was a mistake, I meant to do 

```
df.to_csv(dst_dir)
```

I believe that addresses the issue unless I'm misinterpreting. Please feel free to correct me!

im_v2/common/data/transform/parquet_to_csv.py
DanilYachmenev: Well, you do not need to delete and re-create containing directories or files themselves
So terminologically IIRC you create file 'incrementally', i.e. just create a `.csv` version of the `.pq` file  at the specified location, no need to delete and re-create anything

correct me if I used the term incorrectly
but I hope I conveyed the point

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
...
+import helpers.hs3 as hs3
+import im_v2.common.data.transform.transform_utils as imvcdttrut
+
+_LOG = logging.getLogger(__name__)
+
+def _run(args: argparse.Namespace) -> None:
+    if args.aws_profile is not None:
+        hs3.get_s3fs(args.aws_profile)
+    else:
+        hparser.create_incremental_dir(args.dst_dir, args)
```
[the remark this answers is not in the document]
DanilYachmenev: Well, in the script you specify it is intended to create a destination dir if it does not exist
It does not seem **incorrect** but rather an overkill

IMO we can move the responsibility to pass an existing `dst_dir` to a user and indeed assert it with 
```
hs3.dassert_path_exists(args.dst_dir, aws_profile=aws_profile)
```

This will allow us to check if `dst_dir` is an actual dir and if this is possible to put a file there
This will simplify the interface
wdyt?

devops/compose/docker-compose.yml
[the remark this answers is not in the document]
PomazkinG: for the future reference. I believe GP meant the git branch name, since it's hard to change ok for now

im_v2/common/data/transform/parquet_to_csv.py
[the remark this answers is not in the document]
PomazkinG: I like the idea but probably we should also move `amp/im_v2/common/data/transform/convert_csv_to_pq.py` then. maybe a TODO is fine for now

im_v2/common/data/transform/parquet_to_csv.py
```diff
@@ -0,0 +1,91 @@
...
+        use_relative_paths=True,
+        aws_profile=args.aws_profile,
+    )
+    _LOG.info("Files found at %s:\n%s", args.src_dir, "\n".join(files))
+    for file in files:
+        full_path = os.path.join(args.src_dir, file)
+        _LOG.debug("Converting %s...", full_path)
+
+
+        df = pd.read_parquet(full_path)
```
PomazkinG, line 34: you can even try `from_parquet()` from `helpers.hparquet`

im_v2/common/data/transform/parquet_to_csv.py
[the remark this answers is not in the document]
dchoi127: @DanilYachmenev

in that case should I require the --dst_dir argument? 

Lower in this thread we mentioned putting the --dst_dir as optional with a default value of   None

im_v2/common/data/transform/parquet_to_csv.py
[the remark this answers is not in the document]
dchoi127: Are there any sample parquet files in the sorrentum directory? I can't seem to find any

dchoi127:
Sorry I misclicked and I'm unsure of what I did. I meant to commit my changes and pressed something else. 

https://github.com/sorrentum/sorrentum/pull/267/commits/8dfc4c05c49027cb8c924c5ef07f05f59d4fb536

change is here

DanilYachmenev (commented):
Right direction!

See another review

im_v2/common/data/transform/parquet_to_csv.py
DanilYachmenev: replied in a comment below

im_v2/common/data/transform/parquet_to_csv.py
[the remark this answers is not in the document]
DanilYachmenev: No, there are no
but you can add some dummy files to your local directory for testing
just do not check them in

if you do not have `pq` files, then you can add a `csv` one, convert in to `pq` and then try to convert it back using this script

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+Convert data from Parquet to CSV.
+
+Usage sample:
+> im_v2/common/data/transform/convert_...
+
+Note that the input file can be S3 or local file system, but the destination dir needs to be local.
+
+The code is similar to the conversion from CSV to Parquet performed by
+`im_v2/common/data/transform/convert_csv_to_pq.py`
```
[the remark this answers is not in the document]
PomazkinG: I don't have a strong opinion at the moment. I would leave a TODO like `TODO(x): Consider ....` or `TODO(x): maybe we should ...` just to document the discussion and also communicate that we are not sure what is the decision

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,77 @@
...
+
+def _main(parser: argparse.ArgumentParser) -> None:
+    args = parser.parse_args()
+    hdbg.init_logger(verbosity=args.log_level, use_exec_path=True)
+    hs3.dassert_path_exists(args.dst_dir, aws_profile = args.aws_profile)
+    hs3.dassert_path_exists(args.pq_path)
+    _LOG.info("Files found at %s", args.pq_path)
+    _LOG.debug("Converting %s...", args.pq_path)
+    df = hparquet.from_parquet(args.pq_path)
+    df.to_csv(args.dst_dir)
```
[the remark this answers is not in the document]
dchoi127: My apologies, you're right. That suggestion looks great! Thank you

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,77 @@
...
+import os
+
+import pandas as pd
+
+import helpers.hdbg as hdbg
+import helpers.hparser as hparser
+import helpers.hs3 as hs3
+import helpers.hparquet as hparquet
+
+from pathlib import Path
```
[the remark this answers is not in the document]
dchoi127: my apologies I forgot to run it before committing, got alittle trigger happy.

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python
+
+"""
+Convert data from Parquet to CSV.
+
+Usage sample:
+> im_v2/common/data/transform/convert_...
```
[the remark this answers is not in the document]
dchoi127:
```
im_v2/common/data/transform/convert_pq_to_csv.py --pq_file_path './dataflow/data.parquet' --dst_dir './dataflow'
```

I had a random parquet data file from the market data available to us and put that file in data.parquet. 

Worked as intended.

I also ran the command without the dst_dir flag and worked as intended.

dchoi127:
I've changed the docstring, please let me know if there are any issues with the content/formatting!

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
+"""
+Convert data from Parquet to CSV.
+
+Usage sample:
+> im_v2/common/data/transform/convert_pq_to_csv.py \
+    --pq_file_path '~/Users/d_folder/pq_files/pq_file_to_convert.parquet' \
+    --dst_dir '~/Users/d_folder/csv_files' \
```
DanilYachmenev, line 8: oh lets place some dummy file name and some actual path in the lib
just to make it a bit more familiar for later readers
```
--pq_file_path im_v2/common/data/file_name.pq \
--dst_dir im_v2/common/notebooks \
```

also, you do not need to put backticks or single brackets around file paths and dirs in such cases because you will not be putting them when calling the script

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
...
+    --aws_profile 'ck'
```
DanilYachmenev, line 9: lets remove this from here
`aws_profile` is optional and also if we want to put it here we'd have to use s3 file names and paths

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
+"""
+Convert data from Parquet to CSV.
+
+Usage sample:
```
DanilYachmenev, line 5: `sample` -> `example`

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
...
+> im_v2/common/data/transform/convert_pq_to_csv.py \
+    --pq_file_path '~/Users/d_folder/pq_files/pq_file_to_convert.parquet' \
+    --dst_dir '~/Users/d_folder/csv_files' \
+    --aws_profile 'ck'
+
+Note that dst_dir is optional. If not given, script will save file in \
+same directory that pq_file_path is located in.
```
DanilYachmenev, line 12: we already specify it in `_parse()` at `help` section so lets follow DRY and do not add it here

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
+"""
+Convert data from Parquet to CSV.
```
DanilYachmenev, line 3: -> `Convert file from Parquet to CSV.`

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
...
+Usage sample:
+> im_v2/common/data/transform/convert_pq_to_csv.py \
+    --pq_file_path '~/Users/d_folder/pq_files/pq_file_to_convert.parquet' \
+    --dst_dir '~/Users/d_folder/csv_files' \
+    --aws_profile 'ck'
+
+Note that dst_dir is optional. If not given, script will save file in \
+same directory that pq_file_path is located in.
+
+Note that pq_file_path can be s3 or local. Destination dir can be absolute or relative path
```
DanilYachmenev, line 14: lets remove it from here as well, not as general IMO
we will partially reflect it at `aws_profile` desc, see below

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
...
+        required=True,
+        help="Input Parquet file to convert to CSV format",
+    )
+    parser.add_argument(
+        "--aws_profile",
+        action="store",
+        required=False,
+        default=None,
+        type=str,
+        help="The AWS profile to use for `.aws/credentials` or for env vars",
```
DanilYachmenev, line 46: -> `help="The AWS profile to convert file located at s3 filesystem.`

since we dont really have credentials or env vars here

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    parser.add_argument(
+        "--pq_file_path",
+        action="store",
+        type=str,
+        required=True,
+        help="Input Parquet file to convert to CSV format",
+    )
+    parser.add_argument(
+        "--aws_profile",
```
DanilYachmenev, line 41: lets put it after `dst_dir`
its not crucial, but usually it is better to put params in order of significance (but optional ones are always after required ones ofc)

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
...
+        help="The AWS profile to use for `.aws/credentials` or for env vars",
+    )
+    parser.add_argument(
+        "--dst_dir",
+        action="store",
+        type=str,
+        required=False,
+        default=None,
+        help="Destination dir where to save converted CSV file. \
+        If not specified, uses the same dir where input file is located",
```
DanilYachmenev, line 55: lets format it like
```
help="""
Destination dir where to save converted CSV file.
If not specified, uses the same dir where input file is located`
""",
```
just for consistency

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    parser = hparser.add_verbosity_arg(parser)
+    return parser
+
+def _main(parser: argparse.ArgumentParser) -> None:
+    args = parser.parse_args()
+    dst_dir = args.dst_dir
+    aws_profile = args.aws_profile
+    pq_file_path = args.pq_file_path
+    hdbg.init_logger(use_exec_path= True)
```
DanilYachmenev, line 65: whenever possible, asserts go first in functions, before any actions
in order to catch bugs before doing computations

(and also remove trailing space before `True`)

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,78 @@
...
+    if dst_dir is not None:
+        hs3.dassert_path_exists(dst_dir, aws_profile = aws_profile)
+    hs3.dassert_path_exists(pq_file_path, aws_profile = aws_profile)
+    df = hparque.from_parquet(pq_file_path, aws_profile = aws_profile)
+    if not dst_dir:
+        dst_dir = os.path.dirname(pq_file_path)
```
DanilYachmenev, line 71: Here let's reorder
in general if you work with a param, try to do it in 1 block if possible

in this case interaction with `dst_dir` is unnecessarily scattered
so lets reorder and comment sections like
```
hs3.dassert_path_exists(pq_file_path, aws_profile=aws_profile)
if not dst_dir:
    dst_dir = os.path.dirname(pq_file_path)
hs3.dassert_path_exists(dst_dir, aws_profile=aws_profile)
# Make a path to CSV file.
file_name = os.path.basename(pq_file_path).split(".")[-2]
csv_file_path = os.path.join(dst_dir, file_name + ".csv")
# Convert the file.
df = hparque.from_parquet(pq_file_path, aws_profile=aws_profile)
_LOG.debug("Saving CSV file at %s", csv_file_path)
df.to_csv(csv_file_path)
```

also small note - no need for spaces when specifying optional params inside functions like `aws_profile=aws_profile`

DanilYachmenev (commented):
Very very close, nice!

Just formatting nits

im_v2/common/data/transform/convert_pq_to_csv.py
dchoi127: @DanilYachmenev 

Should the asserts be before lines 62-65?

Or right after line 64?

DanilYachmenev: Actually lets put `hdbg.init_logger(use_exec_path=True)` in `_parse()` as the first line

so it will go before line 30 in the curernt code

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,80 @@
...
+def _main(parser: argparse.ArgumentParser) -> None:
+    args = parser.parse_args()
+    dst_dir = args.dst_dir
+    aws_profile = args.aws_profile
+    pq_file_path = args.pq_file_path
+    hs3.dassert_path_exists(pq_file_path, aws_profile=aws_profile)
+    if not dst_dir:
+        dst_dir = os.path.dirname(pq_file_path)
+    hs3.dassert_path_exists(dst_dir, aws_profile=aws_profile)
+    # Make a path to CSV file
```
DanilYachmenev, line 70: Always finish comments with a dot `.`

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,80 @@
...
+        If not specified, uses the same dir where input file is located
+        """,
+    )
+    parser.add_argument(
+        "--aws_profile",
+        action="store",
+        required=False,
+        default=None,
+        type=str,
+        help="The AWS profile to convert file located at s3 filesystem.",
```
DanilYachmenev, line 55: And in case of help comments, lets not put them
just a habitual ways to do things, you'll get used to it

Ok to leave for `dst_dir` since here are 2 separate sentences

im_v2/common/data/transform/convert_pq_to_csv.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+Convert file from Parquet to CSV.
+
+Usage example:
+> im_v2/common/data/transform/convert_pq_to_csv.py \
+    --pq_file_path im_v2/common/data/file_name.pq \
+    --dst_dir im_v2/common/notebooks
+
+
+Note that pq_file_path can be s3 or local. Destination dir can be absolute or relative path.
```
DanilYachmenev, line 11: Lets reformat it a bit and place before the example
and also - avoid using param and func names in comments and docstring in order to avoid updating them in case of renaming
so
```
"""
Convert file from Parquet to CSV.

- Path to Parquet file can be s3 or local
- Destination dir should belong to the same filesystem as Parquet file (e.g., both at s3 or both at local)
- Destination dir can be absolute or relative path

Usage example:
...
```

DanilYachmenev (commented):
Neat! Just a couple of nits

And the last thing - now we need to check that data is being saved and converted correctly and nothing is lost
Nothing too complicated and nothing to check in:
1) Call the script on your local test files so you have an original `.pq` file and converted `.csv` file
2) open a Jupyter notebook and read both files there using `from_parquet()` and `pd.read_csv()`
3) Check that outputs are the same (you can simply do `df1.equals(df2)`)
4) post here only a screenshot that confirms that script works correctly and data is not being corrupted

and then ready to merge for me

dchoi127:
Screenshot 2023-05-31 at 3 26 08 PM


Reasons why the two dfs are not immediately equal:
1. Dtypes do not match (specifically for datetimes)
2. There minor naming changes that pandas does (ie when there are two timestamp columns the second one is named timestamp.1)

You can see that even after these changes 
```
df1.compare(df2)
```
outputs a column name with no values (which is strange because everything is equal including the column names)

Thus I used [assert_frame_equals](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html#:~:text=Check%20that%20left%20and%20right,for%20use%20in%20unit%20tests.) which executes with no errors.

DanilYachmenev:
> assert_frame_equals

Hmm, we definitely have to fix it
but step by step 

1) `assert_frame_equals` might've passed because you've already implemented changes to the mentioned dataframes. Same way `.equals()` could work I guess
Try to apply it on raw dfs and it will probably fail
2) Can you also display dfs `.head(3)`? So I we could see how dataframes look like exactly
3) It seems that you have 2 `"timestamp"` columns in the original dataframe, right? Lets avoid such cases and just make test df that has no duplicate columns
4) what about  `"knowledge_timestamp"` column? does it really need to be changed?

And see, if this wont work with simpler test dataframes, then check in your test notebooks so i could try it out as well and see what the problem for myself

dchoi127:
> > assert_frame_equals
> 
> Hmm, we definitely have to fix it but step by step
> 
> 1. `assert_frame_equals` might've passed because you've already implemented changes to the mentioned dataframes. Same way `.equals()` could work I guess
>    Try to apply it on raw dfs and it will probably fail
> 2. Can you also display dfs `.head(3)`? So I we could see how dataframes look like exactly
> 3. It seems that you have 2 `"timestamp"` columns in the original dataframe, right? Lets avoid such cases and just make test df that has no duplicate columns
> 4. what about  `"knowledge_timestamp"` column? does it really need to be changed?
> 
> And see, if this wont work with simpler test dataframes, then check in your test notebooks so i could try it out as well and see what the problem for myself

Knowledge_timestamp does need to be changed otherwise it fails based on the fact that df1 has timestamp dtype while df2 has nontimestamp dtype.

Below is a screenshot of the df heads directly after being read in.

Screenshot 2023-05-31 at 3 42 59 PM

dchoi127:
I don't believe there is a way to circumvent the issue because there doesn't appear to be a way to preserve dtypes when exporting the df to a csv. Thus when reading in the csv, it infers the type which leads to the inequality.

DanilYachmenev:
@dchoi127 aha, I see now, thx

It also appears that CSV has no native timestamp format, so it complicates retaining formats

@PomazkinG need your wisdom here
wdyt? Should we try to preserve column dtypes or its fine for now to reflect this in docstrings or a TODO and then we can return to it later when the need appears?

Personally, I'm up for the 2nd option and merge this working script

PomazkinG:
> Personally, I'm up for the 2nd option and merge this working script

I would agree. My logic is: if something is not supported by default -> we keep as-is + add this info to the docstrings. Once it becomes a problem we can add a switch `preserve_types`. The rule is: use-cases drive the development.

dchoi127:
Can I squash and merge now? :D

DanilYachmenev:
> Can I squash and merge now? :D

Noo, actually, for now you NEVER Squash and merge yourself)
This will merge suggested changes to the `master` branch and only integrators can do that (Team Leaders also do this but for a code where everything is clear)

So now I'm passing it to them
Wait for the final review (some nits can be requested by integrators as well)
But overall - good job!
You've made a working useful script that follows our coding standarts rn

dchoi127:
> > Can I squash and merge now? :D
> 
> Noo, actually, for now you NEVER Squash and merge yourself) This will merge suggested changes to the `master` branch and only integrators can do that (Team Leaders also do this but for a code where everything is clear)
> 
> btw @gpsaggese I remember we had some kind of a git tool that actually prevents anyone except specified users to merge PRs Should we enable it to avoid unexpected breaks?
> 
> So now I'm passing it to them Wait for the final review (some nits can be requested by integrators as well) But overall - good job! You've made a working useful script that follows our coding standarts rn

I see I see. The button is tempting to press.

Is there another issue I can start working on? I'd love to learn more about arbitrage and how we do that.

DanilYachmenev:
> I see I see. The button is tempting to press.

I knooow, it took a while for we to get a right to press on it but trust me, it worth it)
 
> Is there another issue I can start working on? I'd love to learn more about arbitrage and how we do that.

@samarth9008 do we have some more?

> As a side note: Thank you for help and patience on this issue. There were many things to change

np, the most important thing is to learn
there will be less and less of them, until there is nothing but pure merge

samarth9008:
> @samarth9008 do we have some more?

Not currently, we need to file more. @DanilYachmenev @dchoi127

PR #197, opened 2023-05-04
Sorr task154 refine supply demand functions

DanilYachmenev:
Task #154 

All the follow-ups from https://github.com/sorrentum/sorrentum/pull/177 are implemented here

Most changes are in fact related to switching from using a list of tuples as coordinates to using `pd.Series` as discussed with @gitpaulsmith

Published notebook:
http://172.30.2.44/notebooks/SorrTask154_Create_linear_supply_demand_curves_using_orders.20230504-051410.html

PR #269, opened 2023-05-19
Sorrtask154 supply/demand follow ups

DanilYachmenev:
Implemented requested follow-ups from https://github.com/sorrentum/sorrentum/pull/197

No results have changed  in the notebooks, so can be considered as a test for correctness

However, we still need to implement unit tests for pathological cases and linear curves

For pathological cases the issue the specs are prepared [here](https://docs.google.com/document/d/1kmybx8u6eJsJBYndPb4r3jPcKM1yNpXIHpLJeRKK4l4/edit?disco=AAAAxOm2KWs), let me know if they are OK

Specs for linear curves unit tests are yet unclear for me
What could we check there exactly?
Because IMO  the current notebook examples are not really visible and easy to read
@gitpaulsmith, wdyt?
Could you give a hint on it? We can make a sync on that if needed, so I could make interface and then delegate probably


gpsaggese (commented):
LGTM. But pls double check with @gitpaulsmith

defi/tulip/notebooks/SorrTask139_Create_some_pathological_cases_for_supply_demand_curve.py
```diff
@@ -142,28 +159,28 @@
...
     limit_price_const=limit_price_const,
 )
-supply_orders_df3 = ddacrord.convert_orders_to_dataframe(supply_orders3)
+supply_orders_df3 = dtuimord.convert_orders_to_dataframe(supply_orders3)
 supply_orders_df3
 
 # %%
 type_ = "supply"
-supply_curve3 = ddcrsede.get_supply_demand_discrete_curve(
+supply_curve3 = dtimsude.get_supply_demand_discrete_curve(
```
gitpaulsmith, line 183: Below you should now do
`pd.concat([supply_curve3, demand_curve1], axis=1).plot()`
instead of
```
plt.plot(supply_curve3)
plt.plot(demand_curve1)
plt.show()
```
That way you'll see the labels.

Same goes elsewhere.

gitpaulsmith (commented):
Let's schedule a discussion to talk through the notebooks.

defi/tulip/notebooks/SorrTask139_Create_some_pathological_cases_for_supply_demand_curve.py
DanilYachmenev: I've tried to experiment with `pd.IntervalIndex` and it rather complicates everything than helps...
However, I found an interesting `plt.step` option that is intended to plot exactly what we need
Снимок экрана 2023-05-30 в 00 45 46

It still requires a certain wrapper in order to plot everything correctly
but it seems that it gives us an opportunity to plot make discrete plots while having normal series of data without duplicate indices

nevertheless, I suggest we pause it for now in order to adjust with the recent shift of company focus for the next 2 months
I'll get back to it when other priorities are implemented
@gitpaulsmith is it OK?

gitpaulsmith: Nice--`plt.step` looks like a good solution. Agree that pausing and switching is best, but if the time it would take to fix the functions and plots is no more than an hour or so, I would fix and merge before pausing.

DanilYachmenev: So in [ef2dc28](https://github.com/sorrentum/sorrentum/pull/269/commits/ef2dc280cbbae1a66a4510e90ff7bf72a5637373)
- I could get rid of duplicate indices, so now we do not have some weird dots in discrete curves series representations - only actual breaking points which are much more readable now
- However, their display required much more to plot everything correctly. Had to create a special function for that which extends curve ends - without them we risk to omit intersections on charts

Also, I've tried to use your suggestion with `pd.concat([supply_curve3, demand_curve1], axis=1).plot()` for aggregate curves, but what confused me that labels are kind of skewed IMO
for example, quote token (ETH) is labeling quantity axis which from my understanding is not really correct (since quantities should be represented through base tokens, right?)
So I think if we want to have labeled plots for aggregate curves, we need a small wrapper for `plt` here as ell 
Снимок экрана 2023-06-01 в 02 59 05

So with all this I think that putting data in the right format is more important than a simple approach to plotting
and well, now I've made a function for it anyways)

Let me know if it is all needed at all and what else I can do here to reach a stopping point

gitpaulsmith (approved):

PR #281, opened 2023-05-25
Updated the docs folder with the markdown files

Aniket-Parlikar:
Added the Docker_guide.md and Jupyter_NB_best _practices files to the docs folder


samarth9008:
@Aniket-Parlikar Whenever you are done with your work, add @DanilYachmenev as you reviewers. Please go through the onboarding docs, so that you get an idea of naming conventions. If needed, refer to other markdown docs in the same folder to get an idea on how to do it.

Issue #261, opened 2023-05-18
Create matching service


tamriq:
1) collect all matching code to the python service that will extract the orders from specific swap from the database
Right now the tulip_orders DB has the following schema: 
```
 id                  | bigint                   |           | not null | nextval('tulip_orders_id_seq'::regclass)
 swap_pair_id        | integer                  |           | not null |
 swap_id             | bigint                   |           | not null |
 order_id            | bigint                   |           | not null |
 order_direction     | character varying(255)   |           |          |
 basetoken           | character varying(255)   |           | not null |
 querytoken          | character varying(255)   |           | not null |
 depositaddress      | character varying(255)   |           | not null |
 senderaddress       | character varying(255)   |           | not null |
 timestamp           | bigint                   |           | not null |
 amount              | numeric                  |           | not null |
 limitprice          | numeric                  |           | not null |
 knowledge_timestamp | timestamp with time zone |           | not null | now()
```
So for each matching round we need to extract the orders that belong to the same `swap_id` and match them.

2) The result of matching we'll send back to Tulip contract, so let's collect them in dicts like [Transfer struct](https://github.com/sorrentum/sorrentum/blob/c165f451f8d95e24c03a6c11bc07d3c558cdbc9d/defi/tulip/contracts/DaoCross.sol#LL33C1-L41C1) we used in 0.1 and 0.2 versions

DanilYachmenev:
@tamriq should we close it or move to background?

tamriq:
@DanilYachmenev let's close

PR #290, opened 2023-05-29
SorrTask261: Create matching service

DanilYachmenev:
#261 

Added a draft of the function to extract matching orders

Now I want to check it.
For this i need `connection` to the DB
@tamriq could you help me with this one pls


defi/tulip/implementation/order_matching.py
```diff
@@ -187,3 +188,54 @@ def get_equivalent_order(
...
+        matching_orders = cursor.fetchall()
+    # Put the matching orders in Solidity code compatible format.
+    transfers = []
+    for order in matching_orders:
+        base_token = order[5]
+        amount = order[10]
+        sender_address = order[8]
+        deposit_address = order[7]
+        transfer = {
+            "token": base_token,
```
DanilYachmenev, line 234: should we pass base or quote token here?

defi/tulip/implementation/order_matching.py
```diff
@@ -187,3 +188,54 @@ def get_equivalent_order(
         order.wallet_address,
     )
     return order
+
+
+def extract_matching_orders(
```
tamriq, line 193: I don't think we need to change the existing order_matching.py file. Instead we want to create a separate service, e.g. here is the new indexer service I've just merged: https://github.com/sorrentum/sorrentum/tree/master/defi/tulip/services/indexer
You need to create a new directory (name it `matching`) in `tulip/services` and a new python file (you can look at [events_indexer.py](https://github.com/sorrentum/sorrentum/blob/master/defi/tulip/services/indexer/events_indexer.py) structure).
The idea is create an independent service that will take the the orders from DB by swap_id, convert each order to `Order` type and put it to `match_orders`. On this stage it's ok to copy paste `match_orders` code from order_matching.py.

defi/tulip/implementation/order_matching.py
```diff
@@ -187,3 +188,54 @@ def get_equivalent_order(
...
+        base_token = order[5]
+        amount = order[10]
+        sender_address = order[8]
+        deposit_address = order[7]
+        transfer = {
+            "token": base_token,
+            "amount": amount,
+            "from": sender_address,
+            "to": deposit_address,
+            "isReturn": True,
```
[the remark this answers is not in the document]
tamriq: In original Transfer struct the idea of "isReturn" field was to separate the transfers that fulfil the orders from the return transfers that send the matching leftovers back to the user. In fact Transfer entity should be the output from  match_orders.
 E.g. in https://github.com/sorrentum/sorrentum/blob/066b7d52c37f6fa2aa5ef0e694cca4e4000797ef/defi/tulip/implementation/order_matching.py#LL142C8-L142C8 we can have some orders left in buy_heap and sell_heap and after the matching is done, we need to convert these unused buy and sell orders to return transfer. E.g. if user wanted to sell 5 tokens but in fact sold only 3, as a result we get two Transfers:
1) that sends the user ethers for 3 matched tokens 
2) that sends 2 unmatched tokens back

Transfer 1) is created in https://github.com/sorrentum/sorrentum/blob/066b7d52c37f6fa2aa5ef0e694cca4e4000797ef/defi/tulip/implementation/order_matching.py#L120
To create Transfer 2) we need to write the code that processes the orders left in buy_heap and sell_heap after the matching is done

defi/tulip/implementation/order_matching.py
DanilYachmenev: Well, it seems like we need to refactor `defi/tulip/implementation/` first, copying the code to the service still involves some changes that need to be discussed closely and separately

let's maybe converge on the interface here and connect it with the current implementation of `match_orders()` code

I can file follow-up bugs and add todos

defi/tulip/services/matching/order_matcher.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def main() -> None:
+    # Get DB connection.
+    db_connection = _get_connection_from_env_vars()
+    # Get global parameters.
+    swap_id = os.environ.get("SWAP_ID")
+    clearing_price = os.environ.get("CLEARING_PRICE")
+    base_token = os.environ.get("BASE_TOKEN")
+    quote_token = os.environ.get("QUOTE_TOKEN")
```
DanilYachmenev, line 100: Should these params be passed like that?

IMO we can at least extract base and quote tokens from DB, but only if  all the matching orders have the same tokens
Can leave a todo for now

defi/tulip/services/matching/order_matcher.py
```diff
@@ -0,0 +1,111 @@
...
+    # Get matching orders from DB.
+    matching_orders = _extract_matching_orders(swap_id, db_connection)
+    # Match orders and get implemented transfers.
+    transfer_df = dtimorma.match_orders(
+        matching_orders, clearing_price, base_token, quote_token
+    )
+    # TODO(Dan): Add next steps.
```
DanilYachmenev, line 107: Not sure where these transfers should go

defi/tulip/services/matching/order_matcher.py
tamriq: In the current version of Tulip contract the orders from the same swap_id have the same base / quote token pair, so we can rely on the data from the database

defi/tulip/services/matching/order_matcher.py
tamriq: For now let's create an empty function `process_transfers` that will take take tulip_address and and transfers as params. And I'll add the web3 logic that'll execute the payments using the functions from Tulip contract

defi/tulip/implementation/order_matching.py
tamriq: Sg

DanilYachmenev:
So after [50fe0fe](https://github.com/sorrentum/sorrentum/pull/290/commits/50fe0fe99b48f3646de5eb186ce613d5bfe942b2) this is how it works with the test data

Снимок экрана 2023-05-30 в 18 00 31

PR #292, opened 2023-05-29
Markdown Linter

samarth9008:


dev_scripts/lint_md.sh
```diff
@@ -1,7 +1,61 @@
-#npm install --save-dev --save-exact prettier
+#!/usr/bin/env bash
+# Markdown linter using prettier.
+#
+# Usage: 
+# npx prettier --parser markdown \
+#   --prose-wrap always \
+#   --write \
+#   --tab-width 2 \
+#   file1.md file2.md ....
```
DanilYachmenev, line 9: I tried it with just like `> dev_scripts/lint_md.sh file1.md` and it worked, so lets update with this easier way is I'm correct
also not sure if all the params specified here are needed

dev_scripts/lint_md.sh
```diff
@@ -1,7 +1,61 @@
...
+
+USER="$(id -u $(logname)):$(id -g $(logname))"
+WORKDIR="$(realpath .)"
+MOUNT="type=bind,source=${WORKDIR},target=${WORKDIR}"
+
+# Append code to initialize the linter.
+cat >./run_linter.sh <<EOF
+
+npx prettier --parser markdown --prose-wrap always --tab-width 2 --write $@
+
```
DanilYachmenev, line 48: is there a way to add a TOH to the processed files?
I tried wit `npx markdown-toc -i $@` but nothing really hapens

DanilYachmenev (commented):
Very cool, it works!
Although, I'm not a very specialist in `.sh` files, so directing it to integrators for a closer look
For now just update the usage example

@gpsaggese with passing it to you, have some questions about what is going to be the process with this script
Now we have 2 linters:
    - Major linter that we all use
        - It is doing more work (adding TOH for example)
        - It is referenced in all the docs
    - This one just for `.md` files
        - Does just basic a formatting
        - Not referenced anywhere
Basically, my question is - what should be the process with this one?
Should we recommend to use it for `.md` instead of a general one in Documentation about guidelines?

dev_scripts/lint_md.sh
samarth9008: What is TOH?

samarth9008: markdown-toc is not added. If needed I can configure it.

DanilYachmenev: Meant TOC, typo

Yes, could you please experiment with adding TOC with your script?
this way we can completely replace the use of general linter with this script and then add here a functionality that is specific of Markdown files

Lets just not spend more than ~30 min on it
if it does not work - its ok, we can file a separate P1 bug  and think about it there

samarth9008: Added TOC functionality. Need to add `<!-- toc -->;` comment at the top of the file or the place where you want to inject the TOC.

dev_scripts/lint_md.sh
samarth9008: I basically took the params from the previous version of the file.

dev_scripts/lint_md.sh
```diff
@@ -1,7 +1,67 @@
-#npm install --save-dev --save-exact prettier
+#!/usr/bin/env bash
+# Markdown linter using prettier.
+#
+# Usage:
```
gpsaggese, line 4: IMO the comment for the user of the container is

```
Lint a markdown file and update the table of content:

Usage
> dev_scripts/lint_md.sh file1.md ... fileN.md

To generate a table of content, add a comment `<!-- toc -->` at the top of the markdown file.
```

Personally I wouldn't add info about npx since those are implementation details that are better documented by the code itself (see DRY)

dev_scripts/lint_md.sh
```diff
@@ -1,7 +1,67 @@
...
+
+# Install Prettier.
+RUN npm install -g prettier && \
+    npm install -g markdown-toc
+
+RUN npx prettier -v
+
+EOF
+
+docker build -f $TMP_FILENAME -t markdownlint .
```
gpsaggese, line 43: Good work. 

Only thing let's call create an env var

DOCKER_CONTAINER_NAME="sorrentum_mdlint"

so we know that it's one of our containers.

gpsaggese (commented):
Good work. One tweak and merge.

Also we need to add info to the markdown author document about this script now.

DanilYachmenev (commented):
LGTM
I'll update info in Documentation about guidelines myself after the merge

gpsaggese (commented):
Merge for me

PR #289, opened 2023-05-28
Sorr task64 refine first review process doc

DanilYachmenev:
#64 

Refined First review doc with some additional processes that we agree to be good practise

PR #295, opened 2023-05-30
SorrTask64: implement First review doc follow-ups

DanilYachmenev:
Implemented here follow-ups from https://github.com/sorrentum/sorrentum/pull/289/files


DanilYachmenev:
Merging to proceed improving the doc
will implement comments in follow-up if any

PR #299, opened 2023-05-30
SorrTask279_add_tulip_diagram

tamriq:
Add Tulip v0.3 diagram to use it in README


gpsaggese:
Sounds good. But next time let's use a tool like Mermaid or C4 so we can have a markdown description that gets rendered

tamriq:
Sure! This is more like a fast temporal solution not to forget the plan while the project is on pause

PR #301, opened 2023-05-31
SorrIssue_Update_Linter_Instructions

samarth9008:


PomazkinG (commented):
I am thinking maybe it's enough to pull the `dev_tools` image once, I mentioned it in https://github.com/sorrentum/sorrentum/pull/303

In any case it's a paranoia thing because the image is pulled under the hood of `i linf -f ...` every time one runs it.

I'd say this PR is redundant, but good job keeping the docs updated

Issue #64, opened 2023-03-21
DOCS - Publish documentation v0.1


gpsaggese:
We can start from our coding conventions 

P1, Let's chat about it

DanilYachmenev:
@gpsaggese

So, after discussion with @tamriq,

At first we need to understand:
1. What documents do we want to publish?
    - Select docs from [Gdrive](https://drive.google.com/drive/u/0/folders/1zawE6IEBDpWLTbpK-03z75f5pu_T9Jba)
    - Set order to split on PRs
2. Where do we want to publish the docs?
    - Should it be a Gdrive or a public Git repo?

We can keep chat about it here and then discuss closely on a Wed call if needed

gpsaggese:
> > We can create a dir Sorrentum and move files one by one. But for each file there needs to be a process like:
> 
> @gpsaggese, all clear, but just to clarify - do you mean to create Sorrentum folder at Gdrive and gradually move reviewed and refined docs there, right?

Correct. We can save them here https://drive.google.com/drive/u/1/folders/1h5jKCiy_jHe2lAIF8HKK-Ah-_Va-xdXh

> 
> And actually we can convert this issue to an EPIC since it will take some time for sure and is very specific. However, I do not really see an option to do it in this repo

Yes, we don't have ZenHub here. I'm ok with just having a single issue (like this one) and then we work on one doc at the time (review, move, delete the old one using the checklist).

The important thing is to do one doc at the time (and not brutal move) and review it for correctness.

DanilYachmenev:
At certain point let's agree on the next docs to publish, since some docs on the [Gdrive](https://drive.google.com/drive/u/0/folders/1sJDqCjM1Q_nq8diyZDiWO8mVBQW5Wg_X) are either not finished, or may be obsolete

For now I suggest to move the docs that are basically finished and contain useful basic info:
- [ ] [Git - Workflows and best practices](https://docs.google.com/document/d/1zahC8uDnFGYRSgkBrQRgg3W3ZmDjZZJj6yln6YeuHq4/edit#heading=h.c2cuxz4yki3)
- [ ] [GitHub / ZenHub workflows](https://docs.google.com/document/d/1M8weNOzMnS7vdMPzEyQ4g7OsCUXXsCn60qzPWsA7SlQ/edit#heading=h.usgoh8klm0ts)
[...]

PR #306, opened 2023-06-02
Sorr task64 update documentation guidelines

DanilYachmenev:
#64 

Updated documentation guidelines with more specific recommendations and markdown linter instructions

note that I decided to not put info about `replace_text` script here, will do it in a coding style doc update since this script is hardly useful now with the new and working markdown linter

FYI @samarth9008


DanilYachmenev:
merging to proceed and reference some added info for students
will implement changes in follow-up if any requested

Issue #244, opened 2023-05-12
Write unit tests to confirm that the timestamp representation for OHLCV data corresponds to what we expect


samarth9008:
Generally we avoid interacting with 3rd party in unit test, that’s why we compromise by writing this as a superslow.

With this test we can keep updating the CCXT without worrying about silently changing the data semantics

In practice

- Let’s start with a test for binance spot and futures
- Pick two pairs (e.g. BTC_USDT, ETH_USDT), download ~5 minutes from a random time interval, i.e. dec 2022 -> this will be the golden outcome
- Use `CcxtExtractor` class to download the OHLCV  data
- Document which version of CCXT was used to download the data (inside the code is enough)
- The test should download the data and compare to the golden outcome to obtain a perfect match

samarth9008:
@jsmerix Is there any reference to **`CcxtExtractor` class to download the OHLCV data**?

samarth9008:
I have assigned myself temporarily. Once someone takes this issue, I will unassign myself.

jsmerix:
> @jsmerix Is there any reference to CcxtExtractor class to download the OHLCV data?

Here is the class [definition](https://github.com/cryptokaizen/cmamp/blob/master/im_v2/ccxt/data/extract/extractor.py)

yiyunlei:
I have some questions for this task:

1. Which specific version of CCXT should I use to generate the golden outcome data?

2. In the process of comparing the newly downloaded OHLCV data with the golden outcome, am I correct in understanding that all data points—not just the timestamp—need to perfectly match?

jsmerix:
> Which specific version of CCXT should I use to generate the golden outcome data?

The one that is currently installed in the dev container, in this case it's `2.1,80`
```
>>> import ccxt
>>> ccxt.__version__
'2.1.80'
```

> In the process of comparing the newly downloaded OHLCV data with the golden outcome, am I correct in understanding that all data points—not just the timestamp—need to perfectly match?

For each timestamp, all fields (open, high, low, close, volume) need to match exactly between golden and actual (the currently downloaded data)

yiyunlei:
When I tried to use `CcxtExtractor` class to download the OHLCV data, I got an ExchangeError and it shows that "Service unavailable from a restricted location". (https://api.binance.com/api/v3/exchangeInfo)
[...]

PR #317, opened 2023-06-08
SorrIssue244 CCXT timestamp representation unit test

yiyunlei:
This is for issue #244.
I finished the unit test for `currency_pairs = ["BTC/USDT"]`. 
Should I add a similar unit test for `currency_pairs = ["ETH/USDT"]`?


im_v2/ccxt/data/extract/test/test_ccxt_version.py
```diff
@@ -0,0 +1,52 @@
...
+import pandas as pd
+
+import helpers.hpandas as hpandas
+import helpers.hunit_test as hunitest
+import im_v2.ccxt.data.extract.extractor as imvcdexex
+
+_LOG = logging.getLogger(__name__)
+
+
+class TestCcxtVersion1(hunitest.TestCase):
```
DanilYachmenev, line 14: Overall the test class name should reflect what class or function it tests
In this case it is `CcxtExtractor` so I'd actually suggest you to not create a new test file and class here but move the test method to `TestCcxtExtractor1` in `im_v2/ccxt/data/extract/test/test_ccxt_extractor.py`, right after `test_download_ohlcv_websocket_kline_is_not_present()` method

im_v2/ccxt/data/extract/test/test_ccxt_version.py
```diff
@@ -0,0 +1,52 @@
...
+    @umock.patch.object(imvcdexex.hdateti, "get_current_time")
+    def test_download_ohlcv_with_current_ccxt_version1(
```
DanilYachmenev, line 16: lets not attach to ccxt version here to much since the goal is to check timestamp representations

so lets keep the used func name and also express what we check here -> `test_download_ohlcv_timestamp_representation1()`

im_v2/ccxt/data/extract/test/test_ccxt_version.py
```diff
@@ -0,0 +1,52 @@
...
+        self, mock_get_current_time: umock.MagicMock
+    ) -> None:
+        """
+        Run with the current ccxt version to verify that data semantics don't
+        change.
```
DanilYachmenev, line 21:
```
Verify that OHLCV data timestamps are correctly represented.
```

im_v2/ccxt/data/extract/test/test_ccxt_version.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        current_time = "2022-10-21 00:00:00.000000+00:00"
+        mock_get_current_time.return_value = current_time
+        # Prepare expected output.
+        # Expected output: This is downloaded data with the CCXT version 2.1.80.
```
DanilYachmenev, line 26: no need for this line one IMO, lets not be to specific

im_v2/ccxt/data/extract/test/test_ccxt_version.py
```diff
@@ -0,0 +1,52 @@
...
+        expected_output = r"""
+        timestamp    open    high    low    close    volume    end_download_timestamp
+        0    1666224060000    19120.48    19123.30    19114.05    19114.05    0.483268    2022-10-21 00:00:00.000000+00:00
+        1    1666224120000    19117.81    19128.66    19117.81    19128.66    0.813596    2022-10-21 00:00:00.000000+00:00
+        2    1666224180000    19128.27    19128.27    19115.57    19122.26    1.800264    2022-10-21 00:00:00.000000+00:00
+        3    1666224240000    19118.08    19122.31    19115.23    19115.23    0.482966    2022-10-21 00:00:00.000000+00:00
+        4    1666224300000    19120.87    19120.87    19113.28    19113.28    0.434573    2022-10-21 00:00:00.000000+00:00
+        """
+        # Initialize class.
+        # Using Binanceus API because Binance API is not accessible.
```
DanilYachmenev, line 36: But this one is a good comment, gj!
lets just rename `Binanceus` -> `Binance.US` for clarity

im_v2/ccxt/data/extract/test/test_ccxt_version.py
```diff
@@ -0,0 +1,52 @@
...
+        exchange_class = imvcdexex.CcxtExtractor("binanceus", "spot")
+        exchange_class.currency_pairs = ["BTC/USDT"]
```
DanilYachmenev, line 38: Can we simply add another currency pair here like `"ETH/USDT"` (and update the goldens ofc)?

DanilYachmenev (commented):
Very good!!
The test is working and passing
And I see that you really read docs and code, gj!

See the suggestions, mostly comments and placement

jsmerix (approved):
Nicely done, make sure to run linter if you haven't already and we can merge.

yiyunlei:
> Very good!! The test is working and passing And I see that you really read docs and code, gj!
> 
> See the suggestions, mostly comments and placement

Thank you!!
I'm really happy that I got praise for my work！

Based on your feedback, I updated the code and encountered a bug.
I left a comment in the issue #244

yiyunlei:
> Nicely done, make sure to run linter if you haven't already and we can merge.

Yes, I have run the linter.

im_v2/ccxt/data/extract/test/test_ccxt_version.py
```diff
@@ -0,0 +1,58 @@
...
+import helpers.hunit_test as hunitest
+import im_v2.ccxt.data.extract.extractor as imvcdexex
+
+_LOG = logging.getLogger(__name__)
+
+
+@pytest.mark.skipif(
+    not henv.execute_repo_config_code("is_CK_S3_available()"),
+    reason="Run only if CK S3 is available",
+)
```
jsmerix, line 21: Your test does not need CK S3 so let's remove this decorator

jsmerix (changes requested):
Noticed one small detail and let's also make sure the test is only present in one file, by the way it does not seems that the problem you described in https://github.com/sorrentum/sorrentum/issues/244#issuecomment-1584901453 is reproduced in github actions since the tests are passing

yiyunlei:
> Noticed one small detail and let's also make sure the test is only present in one file, by the way it does not seems that the problem you described in [#244 (comment)](https://github.com/sorrentum/sorrentum/issues/244#issuecomment-1584901453) is reproduced in github actions since the tests are passing

I copy-pasted the test to the file `test_ccxt_extractor.py` (it's original in the file `test_ccxt_version.py`) based on the feedback from @DanilYachmenev, I also think it's better to put it in the `test_ccxt_extractor.py`

image

So I keep the test in both files now for finding the bug and will delete `test_ccxt_version.py` once I fix the bug.

I run the command `pytest im_v2/ccxt/data/extract/test/test_ccxt_extractor.py::TestCcxtExtractor1::test_download_ohlcv_timestamp_representation1` in the docker and it failed. Could you try this?

Also, how can I run tests in GitHub actions? I didn't find it in docs.

im_v2/ccxt/data/extract/test/test_ccxt_version.py
yiyunlei: I added this because I want to move the ccxt unit test to `test_ccxt_extractor.py` without any bugs. Since `test_ccxt_extractor.py` already has this decorator, I added it there and tried to locate the bug. In the end, I will only keep the test in `test_ccxt_extractor.py`.

DanilYachmenev:
> So I keep the test in both files now for finding the bug and will delete `test_ccxt_version.py` once I fix the bug.

Oh, actually no need to keep both even for now
Its OK to implement all the changes in the old lace and then move all the working code to another directly if you work inside 1 PR
All your previous commit versions are saved in specific commits, like [f16cf73](https://github.com/sorrentum/sorrentum/pull/317/commits/f16cf7387b8a1aa82a89958749e081088596c422)
Снимок экрана 2023-06-10 в 12 58 06

Since your code is already pretty neat, lets remove `test_ccxt_version.py`
In case we need to revert it, you can always access with in the previous commit versions

> I run the command `pytest im_v2/ccxt/data/extract/test/test_ccxt_extractor.py::TestCcxtExtractor1::test_download_ohlcv_timestamp_representation1` in the docker and it failed. Could you try this?

Could you post error logs here pls? So we could we what is the problem?

> Also, how can I run tests in GitHub actions? I didn't find it in docs.

In order to run GitHub actions tests you just need to convert a draft to a PR that is ready for review (right button)
They are not running for drafts but for PRs they will run automatically 
Снимок экрана 2023-06-10 в 13 01 25

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -234,6 +234,43 @@ def test_download_ohlcv_websocket_kline_is_not_present(self) -> None:
...
+        0    1666224060000    19120.48    19123.30    19114.05    19114.05    0.483268    2022-10-21 00:00:00.000000+00:00
+        1    1666224120000    19117.81    19128.66    19117.81    19128.66    0.813596    2022-10-21 00:00:00.000000+00:00
+        2    1666224180000    19128.27    19128.27    19115.57    19122.26    1.800264    2022-10-21 00:00:00.000000+00:00
+        3    1666224240000    19118.08    19122.31    19115.23    19115.23    0.482966    2022-10-21 00:00:00.000000+00:00
+        4    1666224300000    19120.87    19120.87    19113.28    19113.28    0.434573    2022-10-21 00:00:00.000000+00:00
+        """
+        # Initialize class.
+        # Using Binance.US API because Binance API is not accessible.
+        exchange_class = imvcdexex.CcxtExtractor("binanceus", "spot")
+        exchange_class.currency_pairs = ["BTC/USDT"]
```
DanilYachmenev, line 258: formatting nit
if some value is being repeatedly used in the code, set is as a param, so in case it changes, you do not forget to update all the calls

so here
```
exchange_id = "binanceus"
currency_pair = "BTC/USDT"
exchange_class = imvcdexex.CcxtExtractor(exchange_id "spot")
exchange_class.currency_pairs = [currency_pair]
...
ccxt_data = exchange_class._download_ohlcv(
            exchange_id=exchange_id,
            currency_pair=currency_pair,
...
```

im_v2/ccxt/data/extract/test/test_ccxt_version.py
DanilYachmenev: Actually, does not seem that this suggestion is reasonable, since `_download_ohlcv` works only with 1 curr pair and `expected_output` signature just has no a column for currency pair - so mixing the signatures would be a mess

therefore, the test for 1 curr pair is enough IMO

DanilYachmenev (commented):
almost there
lets do these nits, fix the error and then pass to GP

yiyunlei:
Error logs:

The test `test_download_ohlcv_timestamp_representation1` can run successfully in the new file `test_ccxt_version.py`, but after I moved it to the `text_ccxt_extractor.py`, I got an unexpected bug.

- Test result when running the test in `test_ccxt_version.py`:
`pytest im_v2/ccxt/data/extract/test/test_ccxt_version.py::TestCcxtVersion1::test_download_ohlcv_timestamp_representation1`
```shell
im_v2/ccxt/data/extract/test/test_ccxt_version.py::TestCcxtVersion1::test_download_ohlcv_timestamp_representation1 (1.22 s) PASSED                                                           [100%]

=================================================================================== slowest 3 durations ===================================================================================
1.26s call     im_v2/ccxt/data/extract/test/test_ccxt_version.py::TestCcxtVersion1::test_download_ohlcv_timestamp_representation1
0.01s setup    im_v2/ccxt/data/extract/test/test_ccxt_version.py::TestCcxtVersion1::test_download_ohlcv_timestamp_representation1
0.00s teardown im_v2/ccxt/data/extract/test/test_ccxt_version.py::TestCcxtVersion1::test_download_ohlcv_timestamp_representation1
```

- Test result when running the test in `test_ccxt_extractor.py`:
`
pytest im_v2/ccxt/data/extract/test/test_ccxt_extractor.py::TestCcxtExtractor1::test_download_ohlcv_timestamp_representation1
`
```shell
======================================================================================== FAILURES =========================================================================================
____________________________________________________________ TestCcxtExtractor1.test_download_ohlcv_timestamp_representation1 _____________________________________________________________
  File "/app/im_v2/ccxt/data/extract/test/test_ccxt_extractor.py", line 262, in test_download_ohlcv_timestamp_representation1
    ccxt_data = exchange_class._download_ohlcv(
  File "/app/im_v2/ccxt/data/extract/extractor.py", line 187, in _download_ohlcv
    range(
TypeError: only integer scalar arrays can be converted to a scalar index
=================================================================================== slowest 3 durations ===================================================================================
0.15s call     im_v2/ccxt/data/extract/test/test_ccxt_extractor.py::TestCcxtExtractor1::test_download_ohlcv_timestamp_representation1
0.01s setup    im_v2/ccxt/data/extract/test/test_ccxt_extractor.py::TestCcxtExtractor1::test_download_ohlcv_timestamp_representation1
0.00s teardown im_v2/ccxt/data/extract/test/test_ccxt_extractor.py::TestCcxtExtractor1::test_download_ohlcv_timestamp_representation1
================================================================================= short test summary info =================================================================================
FAILED im_v2/ccxt/data/extract/test/test_ccxt_extractor.py::TestCcxtExtractor1::test_download_ohlcv_timestamp_representation1 - TypeError: only integer scalar arrays can be converted t...
```

The related code in `extrator.py`:
```python
        # Convert datetime into ms.
        start_timestamp = start_timestamp.asm8.astype(int) // 1000000
        end_timestamp = end_timestamp.asm8.astype(int) // 1000000
        duration = self._sync_exchange.parse_timeframe("1m") * 1000
        all_bars = []
        # Iterate over the time period.
        # Note: the iteration goes from start date to end date in milliseconds,
        # with the step defined by `bar_per_iteration` parameter.
        # Because of this, the output can go slightly over the end date.
        for t in tqdm.tqdm(
            range(
                start_timestamp,
                end_timestamp + duration,
                duration * bar_per_iteration,
            )
        ):
```
@DanilYachmenev

yiyunlei:
I still get this error for now when I run this unit test in the docker, and I can't test in the GitHub actions since GitHub actions tests are not available on Sorrentum yet.

DanilYachmenev:
> Error logs:

This is indeed a complicated error that is related to the mocking approach
I myself actually is not an expert in this

@jsmerix, could you ptal?

I've run the test with some logs added for the params that go to the `range` in `_download_ohlcv()` and the problem is definitely in computing `duration` at line 180:
`duration = self._sync_exchange.parse_timeframe("1m") * 1000`
param logs:
```
(0.02 s) duration <MagicMock name='ccxt.binanceus().parse_timeframe().__mul__()' id='140105328842400'>
start_timestamp 1666224060000
end_timestamp 1666224300000
end_timestamp + duration []
duration * bar_per_iteration <MagicMock name='ccxt.binanceus().parse_timeframe().__mul__().__mul__()' id='140105330215760'>
```


So as you can see, `duration` is not parsed and seems like it requires timeframe parsing, like it is done for `test_download_ohlcv1`:
```
        # Mock a call to ccxt's `parse_timeframe method` called inside `_fetch_ohlcv`.
        with umock.patch.object(
            exchange_class._sync_exchange, "parse_timeframe", create=True
        ) as parse_timeframe_mock:
            parse_timeframe_mock.return_value = 60
            # Run.
            ohlcv_data = exchange_class._download_ohlcv(
            ...
``` 

When I simply done the same in the test, all the values turned into int variables
However, an empty df is returned, so I'm questioning what else should be done to download the data correctly in this case
param logs:
```
(0.03 s) duration 60000
start_timestamp 1645660800000
end_timestamp 1645747200000
end_timestamp + duration 1645747260000
duration * bar_per_iteration 30000000
```
output df is empty, test fails:
```
exp = r"""
Empty DataFrame
Columns: [timestamp, open, high, low, close, volume, end_download_timestamp]
Index: []
"""
```

My guess on why the same code now fails is that in `TestCcxtExtractor1` there is mock call to CCXT, but not sure
```
class TestCcxtExtractor1(hunitest.TestCase):
    # Mock calls to external providers.
    ccxt_patch = umock.patch.object(imvcdexex, "ccxt", spec=ccxt)
```

So this is where I could get so far
What else could we do to proceed with debugging and do a fix?

jsmerix:
@yiyunlei I'd put the test into a new test class (new class, same file) and see if it passes, no reason to lose sleep over this.

yiyunlei:
I now understand the 'superslow' in the issue #244 description, as this unit test needs to download data through an API.

So, how can I make this unit test a superslow test? I couldn't find any doc about it. 
I think this will be the final step for this task.

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,46 @@ def test_download_bid_ask_invalid_input1(self) -> None:
         actual = str(fail.value)
         expected = "Currency pair is not present in exchange"
         self.assertIn(expected, actual)
+
+
+class TestCcxtExtractor2(hunitest.TestCase):
+    @umock.patch.object(imvcdexex.hdateti, "get_current_time")
+    def test_download_ohlcv_timestamp_representation1(
```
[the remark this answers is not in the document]
yiyunlei: `pytest` is imported

.gitignore
```diff
@@ -1,4 +1,4 @@
-.DS_Store   # macOS hidden files.
+.DS_Store
```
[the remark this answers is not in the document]
yiyunlei: How about moving the comment to a new line? 
Because comments must start at the beginning of a line and a line with a comment at the end does not work. 
Resource: https://stackoverflow.com/a/8865858/18517831
image

Or should I open a new issue for it?

DanilYachmenev:
> @jsmerix Could you unblock this merge?

merge button is only for GP here)

but the button is being unlocked only when you resolve all the conversations

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,46 @@ def test_download_bid_ask_invalid_input1(self) -> None:
...
+        # Using Binance.US API because Binance API is not accessible.
+        exchange_id = "binanceus"
+        currency_pair = "BTC/USDT"
+        currency_type = "spot"
+        start_timestamp = pd.Timestamp("2022-10-20T00:01:00Z")
+        end_timestamp = pd.Timestamp("2022-10-20T00:05:00Z")
+        # Initialize class.
+        exchange_class = imvcdexex.CcxtExtractor(exchange_id, currency_type)
+        exchange_class.currency_pairs = [currency_pair]
+        # Download real OHLCV data using ccxt.
```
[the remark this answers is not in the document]
DanilYachmenev: No, just in the comments we use it capitalized 
in names everything os ok

.gitignore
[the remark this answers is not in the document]
DanilYachmenev: well, yes, lets move it above the code line then

yiyunlei:
> > @jsmerix Could you unblock this merge?
> 
> merge button is only for GP here)
> 
> but the button is being unlocked only when you resolve all the conversations

Thanks, but there is a requested change from @jsmerix, and I have addressed it.

image
image

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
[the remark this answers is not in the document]
yiyunlei: Oh thanks, I misunderstood.

yiyunlei:
Resolved all the conversations and updated code.

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,47 @@ def test_download_bid_ask_invalid_input1(self) -> None:
...
+        1    1666224120000    19117.81    19128.66    19117.81    19128.66    0.813596    2022-10-21 00:00:00.000000+00:00
+        2    1666224180000    19128.27    19128.27    19115.57    19122.26    1.800264    2022-10-21 00:00:00.000000+00:00
+        3    1666224240000    19118.08    19122.31    19115.23    19115.23    0.482966    2022-10-21 00:00:00.000000+00:00
+        4    1666224300000    19120.87    19120.87    19113.28    19113.28    0.434573    2022-10-21 00:00:00.000000+00:00
+        """
+        # Initialize parameters.
+        # Using Binance.US API because Binance API is not accessible.
+        exchange_id = "binanceus"
+        currency_pair = "BTC/USDT"
+        currency_type = "spot"
```
jsmerix, line 715: nit detail: rename the variable `currency_type -> asset_type` (internal convention)

jsmerix (changes requested):
Let\s add the same test for `ETH/USDT`. Make sure we do not have a naive copy paste solution rather utilize some helper function

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,47 @@ def test_download_bid_ask_invalid_input1(self) -> None:
...
+        mock_get_current_time.return_value = current_time
+        # Prepare expected output.
+        expected_output = r"""
+        timestamp    open    high    low    close    volume    end_download_timestamp
+        0    1666224060000    19120.48    19123.30    19114.05    19114.05    0.483268    2022-10-21 00:00:00.000000+00:00
+        1    1666224120000    19117.81    19128.66    19117.81    19128.66    0.813596    2022-10-21 00:00:00.000000+00:00
+        2    1666224180000    19128.27    19128.27    19115.57    19122.26    1.800264    2022-10-21 00:00:00.000000+00:00
+        3    1666224240000    19118.08    19122.31    19115.23    19115.23    0.482966    2022-10-21 00:00:00.000000+00:00
+        4    1666224300000    19120.87    19120.87    19113.28    19113.28    0.434573    2022-10-21 00:00:00.000000+00:00
+        """
```
DanilYachmenev, line 710: 1 more nit
do
```
# pylint: disable=line-too-long
expected_output = r"""
...
"""
# pylint: enable=line-too-long
```
to please the linter

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
jsmerix: That works as well

yiyunlei:
> 

Finished and tests passed.

yiyunlei:
> Let\s add the same test for `ETH/USDT`. Make sure we do not have a naive copy paste solution rather utilize some helper function

Finished and tests passed.

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,98 @@ def test_download_bid_ask_invalid_input1(self) -> None:
         actual = str(fail.value)
         expected = "Currency pair is not present in exchange"
         self.assertIn(expected, actual)
+
+
+class TestCcxtExtractor2(hunitest.TestCase):
+    @pytest.mark.superslow("~30 seconds.")
+    @umock.patch.object(imvcdexex.hdateti, "get_current_time")
+    def test_download_ohlcv_timestamp_representation1(
```
DanilYachmenev, line 694: avoid nested functions, especially when they are so big - difficult to review and edit

instead do an independent `_helper()` method and use in corresponding tests
see `Test_compute_duration_df` in `helpers/test/test_hpandas.py` for ex

yiyunlei: Finished

yiyunlei:
Used independent _helper() and use two tests for two currency pairs. Also, tests passed.

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,110 @@ def test_download_bid_ask_invalid_input1(self) -> None:
...
+        Verify that OHLCV data timestamps for exchange ID Binance.US and
+        currency pair "BTC/USDT".
+        """
+        # Using Binance.US API because Binance API is not accessible.
+        exchange_id = "binanceus"
+        currency_pair = "BTC/USDT"
+        contract_type = "spot"
+        start_timestamp = pd.Timestamp("2022-10-20T00:01:00Z")
+        end_timestamp = pd.Timestamp("2022-10-20T00:05:00Z")
+        # Prepare expected output
```
DanilYachmenev, line 740: a tiny nit: end comment with dots`.`
check others too

DanilYachmenev (commented):
good job! LGTM

a tiny nit to adhere standarts though, pls do

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
yiyunlei: No problem, I will pay more attention to these details.

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,110 @@ def test_download_bid_ask_invalid_input1(self) -> None:
         actual = str(fail.value)
         expected = "Currency pair is not present in exchange"
         self.assertIn(expected, actual)
+
+
+class TestCcxtExtractor2(hunitest.TestCase):
+    def test_download_ohlcv_timestamp_representation_helper(
```
jsmerix, line 692: nit: private non-test method should start with underscore `_test_download_ohlcv_timestamp_representation_helper`

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,110 @@ def test_download_bid_ask_invalid_input1(self) -> None:
...
+        exchange_id: str,
+        currency_pair: str,
+        contract_type: str,
+        start_timestamp: pd.Timestamp,
+        end_timestamp: pd.Timestamp,
+        expected_output: str,
+        mock_get_current_time: umock.MagicMock,
+    ) -> None:
+        """
+        Verify that OHLCV data timestamps are correctly represented.
```
jsmerix, line 703: Based on the observations of our code base we should improve test case docstrings so it is easier to understand what the test is doing, basically reiterating what the original specs required

In this case:
```
"""
Verify that OHLCV data timestamps are correctly represented.

1. Perform a historical download of OHLCV data for a given time period and currency pair
2. Compare it to previously downloaded data using a specific CCXT version. This test ensures we can detect a change in interface (i.e. timestamp representation in this case) upon upgrading the library version.
"""
```

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
```diff
@@ -686,3 +686,110 @@ def test_download_bid_ask_invalid_input1(self) -> None:
...
+        """
+        # Mock current time.
+        current_time = "2022-10-21 00:00:00.000000+00:00"
+        mock_get_current_time.return_value = current_time
+        expected_output = expected_output.strip()
```
jsmerix, line 708: Was this line necessary? (It might have been, I am just asking)

jsmerix (changes requested):
A few details left, bare with us, the first PRs always have this back-and-forth vibe.

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
yiyunlei: Don't start with an underscore here because I followed the example `helpers/test/test_hpandas.py`. This example doesn't begin with an underscore for a similar helper.
image
Just to make sure which rule to follow.

jsmerix: Sure why not, as I said, it's a nit-picky remark, we can keep it your way, fine by me, let's if GP sneaks in with a decisive opinion.

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
yiyunlei: Actually, not necessary. I'll delete it.

im_v2/ccxt/data/extract/test/test_ccxt_extractor.py
yiyunlei: `_test_download_ohlcv_timestamp_representation_helper` is better IMO, so I'll use this one for now.

yiyunlei:
The linter placed the helper function at the end.

jsmerix (approved):
Ready to merge, well done.

gpsaggese (commented):
Good work

yiyunlei:
@DanilYachmenev @jsmerix Thanks for all the reviews! They helped me a lot and also enhanced my understanding of code style and how to write clean, understandable code.

Issue #319, opened 2023-06-11
Refine `convert_gdoc_to_markdown.sh`


DanilYachmenev:
Follow-up from https://app.asana.com/0/0/1204574767282612/1204799311238928/f

Direct call of `dev_scripts/convert_gdoc_to_markdown.sh` when trying to convert a `.docx` to `.md` fails:
```
dev_scripts/convert_gdoc_to_markdown.sh: line 16: pandoc: command not found
```

we need to  give to `pandoc` the usual treatment of creating a dockerized executable like it is done in `dev_scripts/lint_md.sh`

FYI @samarth9008

DanilYachmenev:
implemented in #320

PR #320, opened 2023-06-11
SorrTask319: Refine convert_gdoc_to_markdown.sh

DanilYachmenev:
#319 

So I've added my perception of how a docker executable can be implemented by looking at `lint_md` script
However, not sure if all is needed or correct, not experienced with this type of scripts
@samarth9008 could you ptal?

Also, added some files that I used the script on so you could see the result and we could update how it works


dev_scripts/convert_gdoc_to_markdown.sh
```diff
@@ -1,23 +1,43 @@
...
-    eval $cmd
+    CMD="pandoc --extract-media $OUT_FIGS -f docx -t markdown -o $OUT_FILE $IN_FILE"
+    docker run --rm -it --workdir "${WORKDIR}" --mount "${MOUNT}" ${DOCKER_CONTAINER_NAME} ${CMD}
 
-    # Move the media.
-    mv $OUT_FIGS/{media/*,}
-    rm -rf $OUT_FIGS/media
+    # # Move the media.
+    # mv $OUT_FIGS/{media/*,}
+    # rm -rf $OUT_FIGS/media
```
DanilYachmenev, line 40: This results to 
```
+ mv 'docs/Sprint_planning_process_figs/media/*' docs/Sprint_planning_process_figs/
mv: cannot stat 'docs/Sprint_planning_process_figs/media/*': No such file or directory
```
tried to bypass but need a direction, seems like a simple thing to do

dev_scripts/convert_gdoc_to_markdown.sh
DanilYachmenev, line 1: should we rename it to `convert_docx_to_markdown.sh` since we're coverting `docx`, not Gdocs here?

dev_scripts/convert_gdoc_to_markdown.sh
```diff
@@ -1,23 +1,43 @@
 #!/bin/bash -xe
-#IN_FILE='/Users/saggese/Downloads/Tools\ -\ PyCharm.docx'
-#OUT_PREFIX="docs/Tools-PyCharm"
-#OUT_PREFIX="defi/papers/sorrentum"
-OUT_PREFIX="docs/DataFlow"
-#OUT_PREFIX="docs/DataPull"
+
+set -eux
+
```
DanilYachmenev, line 4: Definitely need a docstring
Lets add when agree on the code

docs/Sprint_planning_process.md
```diff
@@ -0,0 +1,48 @@
+## Sprints
+
```
DanilYachmenev, line 2: It seems like we can add a feature to the script that adds a TOC automatically

e.g. puts `<!-- toc -->`  on top of the file before running `lint_md`

docs/Sprint_planning_process.md
```diff
@@ -0,0 +1,48 @@
...
+## Epics
+
+- Epics are thematic groups of related issues
+
+- Most epics are created around software components or user workflows (which may
+  cross software components)
+
+- Epics with a high level of activity are versioned and have a well-defined
+  purpose
```
DanilYachmenev, line 38: Should we also remove all these empty lines between 1 bullet list points?
Just an idea for now, not sure how complicated is it for now

dev_scripts/convert_gdoc_to_markdown.sh
gpsaggese: Agreed. I thought the same.

dev_scripts/convert_gdoc_to_markdown.sh
```diff
@@ -1,23 +1,43 @@
...
+# Build the Docker container.
+TMP_FILENAME="/tmp/tmp._convert_docx_to_markdown.Dockerfile"
```
gpsaggese, line 6: let's create the file here to avoid collisions on the shared /tmp

also it's easier to debug and it will be cleaned up

dev_scripts/convert_gdoc_to_markdown.sh
gpsaggese: I think we need to create the dir if it doesn't exist

if [[ !-d docs/Sprint_planning_process_figs/ ]]; then
   mkdir

docs/Sprint_planning_process.md
gpsaggese: We can leave it as a TODO

docs/Sprint_planning_process.md
gpsaggese: I remove the empty lines but it needs to be manual.
We can add a TODO in the shell script
```
# TODO: Find a way to remove empty lines in bullet list.
```

I would leave it as whatever the lint_md does for now

dev_scripts/convert_gdoc_to_markdown.sh
```diff
@@ -1,23 +1,43 @@
 #!/bin/bash -xe
-#IN_FILE='/Users/saggese/Downloads/Tools\ -\ PyCharm.docx'
-#OUT_PREFIX="docs/Tools-PyCharm"
-#OUT_PREFIX="defi/papers/sorrentum"
-OUT_PREFIX="docs/DataFlow"
-#OUT_PREFIX="docs/DataPull"
+
```
gpsaggese, line 2: Let's add a TODO

```
# TODO(gp): @all This script is becoming too complicated and it should be converted in Python. Consider this before adding more complexity.
```

This can be a first issue for a new contributor.
Feel free to file a bug

gpsaggese (commented):
Looking good

dev_scripts/convert_gdoc_to_markdown.sh
DanilYachmenev: Filed #322

dev_scripts/convert_gdoc_to_markdown.sh
DanilYachmenev: The problem was actually for the cases when the doc has no images so `media` folder is not being created
fixed

dev_scripts/convert_gdoc_to_markdown.sh
DanilYachmenev: Created at `/dev_scrpts`
Works as expected, tmp file is being removed

Should we do same for `lint_md.sh`?

PR #331, opened 2023-06-15
Sorrtask298 rename get docker cmd to get docker run cmd

zli00185:


samarth9008:
Does your script work as expected? If so, push those files too where replacement has happened.

zli00185:
The directory path will change if I run it locally. Do I keep the local path or the one from Juypter notebook?

dev_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash -xe
+
+# Change the directory to the root
+# In Juypter note the root is /app
+dir_names="/Users/leosmacbook/Desktop/Sorretum_Intern"
```
samarth9008, line 5: This dir name doesn't work for us as it us local to your system. Update the script accordingly. Take a look at other scripts.

samarth9008:
> The directory path will change if I run it locally. Do I keep the local path or the one from Juypter notebook?

Whats the need of using jupyter notebook in this?

dev_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
DanilYachmenev: And also, I see no reason to use it here at all
in fact in current shape the changes will be done only in the specified dir

read `replace_text.py` to see how params do work

zli00185:
I though we would use the path from docker container since that is same for everyone.

dev_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
samarth9008: @zli00185 When you make any changes according to the comments, please click the Resolve button. When submit a PR add reviewers so that they are notified.

dev_scripts/cleanup_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -xe
+
+curr_file="/amp/dev_scripts/cleanup_scripts dev_scripts/cleanup_scripts"
```
DanilYachmenev, line 3: no need for this one either

dev_scripts/cleanup_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
```diff
@@ -0,0 +1,12 @@
...
+
+replace_text.py \
+    --old "_get_docker_cmd" \
+    --new "_get_docker_run_cmd" \
+    --exclude_dirs "$curr_file"
```
DanilYachmenev, line 8: and no need for this as well

DanilYachmenev (commented):
no need to exclude the script itself

when nit is fixed, lgtm

dev_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
zli00185: okay

samarth9008:
You committed wrong changes. Now your old and new both are same so basically nothing will be replaced. You needed to remove currfiles and exclude dir. After doing that test your script again. If it works then only push it.

samarth9008:
Modify your script correctly. Create a dummy file to test it. If it works, delete the dummy file and only push the new scripts. Now you cannot test on the previous files because they are already replaced.

dev_scripts/cleanup_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/bash -xe
 
-curr_file="/amp/dev_scripts/cleanup_scripts dev_scripts/cleanup_scripts"
 
 replace_text.py \
     --old "_get_docker_run_cmd" \
-    --new "_get_docker_run_cmd" \
```
samarth9008: The `--old` and `--new` has same args. So this is not correct.

dev_scripts/cleanup_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/bash -xe
 
-curr_file="/amp/dev_scripts/cleanup_scripts dev_scripts/cleanup_scripts"
```
samarth9008, line 3: Remove empty space.

samarth9008 (changes requested):

dev_scripts/cleanup_scripts/SorrTask298_rename_get_docker_cmd_to_get_docker_run_cmd.sh
samarth9008: Nothing changed over here.

zli00185: The file itself will get rename as well. So I need the version before running it, right?

samarth9008: Yes.

samarth9008 (approved):

PR #334, opened 2023-06-16
Convert_Type_Hints_gDoc_to_Markdown_#328

hhxjqm:

PR #335, opened 2023-06-17
SorrIssue328 Convert Type Hints gDoc to Markdown

hhxjqm:
Instructions told me that "The entire conversion flow is in dev_scripts/convert_gdoc_to_markdown.sh", but I can't find it, now I know it called ./dev_scripts/convert_docx_to_markdown.sh.

I redo it and here is update

### DIFF
--- a/docs/Python_Type_hints.md
+++ b/docs/Python_Type_hints.md
@@ -1,0 +1,270 @@
+<!--ts-->
+   * [# Why we use type hints](#-why-we-use-type-hints)
+   * [# What to annotate with type hints](#-what-to-annotate-with-type-hints)
+   * [# Conventions](#-conventions)
+      * [## Empty return](#-empty-return)
+      * [## Invoke tasks](#-invoke-tasks)
+      * [## Annotation for kwargs](#-annotation-for-kwargs)
+      * [## Any](#-any)
+      * [## np.array and np.ndarray](#-nparray-and-npndarray)
+   * [# Handling the annoying Incompatible types in assignment](#-handling-the-annoying-incompatible-types-in-assignment)
+   * [# Handling the annoying "None" has no attribute](#-handling-the-annoying-none-has-no-attribute)
+   * [# Disabling mypy errors](#-disabling-mypy-errors)
+   * [# What to do when you don't know what to do](#-what-to-do-when-you-dont-know-what-to-do)
+   * [# Library without types](#-library-without-types)
+   * [# Inferring types using unit tests](#-inferring-types-using-unit-tests)
+
+
+
+<!--te-->
+# \# Why we use type hints
+
+We use Python 3 type hints to:
+
+- Improve documentation
+- Allow mypy to perform static checking of the code, looking for bugs
+- Enforce the type checks at run-time, through automatic assertions (not
+  implemented yet)
+
+# \# What to annotate with type hints
+
+- We expect all new library code (i.e., that is not in a notebook) to have type
+  annotations
+- We annotate the function signature
+- We don't annotate the variables inside a function unless mypy reports that it
+  can't infer the type
+- We strive to get no errors / warnings from the linter, including mypy
+
+# \# Conventions
+
+## \#\# Empty return
+
+Return `-> None` if your function doesn't return
+
+- Pros:
+  - `mypy` checks functions only when there is at least an annotation: so using
+    `-> None` enables mypy to do type checking
+  - It reminds us that we need to use type hints
+- Cons:
+  - `None` is the default value and so it might seem redundant
+
+## \#\# Invoke tasks
+
+For some reason `invoke` does not like type hints, so we
+
+- Omit type hints for `invoke` tasks, i.e. functions with the `@task` decorator
+- Put `\# type: ignore` so that `mypy` does not complain Example:
+  `` \` @task def run_qa_tests( \# type: ignore ctx, stage="dev", version="", ):  ``\`
+
+## \#\# Annotation for `kwargs`
+
+- We use `kwargs: Any` and not `kwargs: Dict[str, Any]`
+- `\*` always binds to a `Tuple`, and `\*\*` always binds to a `Dict[str, Any]`.
+  Because of this restriction, type hints only need you to define the types of
+  the contained arguments. The type checker automatically adds the
+  `Tuple[_, ...]` and `Dict[str, _]` container types.
+- [[Reference article]{.underline}](https://adamj.eu/tech/2021/05/11/python-type-hints-args-and-kwargs/)
+
+## \#\# `Any`
+
+- `Any` type hint = no type hint
+- We try to avoid it everywhere when possible
+
+## \#\# `np.array` and `np.ndarray`
+
+- If you get something like the following lint:
+  `` \` dataflow/core/nodes/sklearn_models.py:537:[amp_mypy] error: Function "numpy.core.multiarray.array" is not valid as a type [valid-type]  ``\`
+  > Then the problem is probably that a parameter that the lint is related to
+  > has been types as `np.array` while it should be typed as `np.ndarray`:
+  >
+  > `x_vals: np.array` -> `x_vals: np.ndarray`
+
+# \# Handling the annoying `Incompatible types in assignment`
+
+- `mypy` assigns a single type to each variable for its entire scope
+- The problem is in common idioms where we use the same variable to store
+  different representations of the same data
+  > ``\`
+  >
+  > output : str = ...
+  >
+  > output = output.split("\\n")
+  >
+  > ...
+  >
+  > \# Process output.
+  >
+  > ...
+  >
+  > output = "\\n".join(output)
+  >
+  > ``\`
+- Unfortunately the proper solution is to use different variables
+  > ``\`
+  >
+  > output : str = ...
+  >
+  > output_as_array = output.split("\\n")
+  >
+  > ...
+  >
+  > \# Process output.
+  >
+  > ...
+  >
+  > output = "\\n".join(output_as_array)
+  >
+  > ``\`
+- Another case could be:
+  > ``\`
+  >
+  > from typing import Optional
+  >
+  > def test_func(arg: bool):
+  >
+  > ...
+  >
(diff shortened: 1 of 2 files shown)


DanilYachmenev:
@hhxjqm pls carefully read https://github.com/sorrentum/sorrentum/blob/master/docs/First_review_process.md#save-reviewer-time 

treat your PR accordingly, use labels and assign team leaders first

DanilYachmenev:
also, should we consider #334 obsolete?

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Invoke tasks
+
+For some reason `invoke` does not like type hints, so we
+
+- Omit type hints for `invoke` tasks, i.e. functions with the `@task` decorator
+- Put `# type: ignore` so that `mypy` does not complain
+
+Example:
+  ```python
```
DanilYachmenev, line 59: nicee

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Annotation for `kwargs`
+
+- We use `kwargs: Any` and not `kwargs: Dict[str, Any]`
+- `*` always binds to a `Tuple`, and `**` always binds to a `Dict[str, Any]`.
+  Because of this restriction, type hints only need you to define the types of
+  the contained arguments. The type checker automatically adds the
+  `Tuple[_, ...]` and `Dict[str, _]` container types.
+- [Reference article](https://adamj.eu/tech/2021/05/11/python-type-hints-args-and-kwargs/)
+## `Any`
```
DanilYachmenev, line 75: maintain empty spaces before and after headers

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,229 @@
...
+  1. Type hint when assigning the model parameter in ctor:
+     ```bash
+     self._model: Optional[sklearn.base.BaseEstimator] = None
+     ```
+  2. Cast a type to the model parameter after asserting that it is not `None`:
+     ```bash
+     hdbg.dassert_is_not(self._model, None) self._model = cast(sklearn.base.BaseEstimator, self._model)
+     ```
+
+# Disabling `mypy` errors
```
DanilYachmenev, line 159: This should be a header format, right?
for some reason displayed as a comment

hhxjqm: It works well in display the rich diff. Still trying figure out.

DanilYachmenev: Well, it looks correct indeed and also reflected in TOH
So IMO its ok

DanilYachmenev (commented):
lgtm!

DanilYachmenev:
@samarth9008 could you ptal, check for some omissions that I could make and check out the unresolved comment?
If everything is OK, passing to GP

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+  - Omit type hints for `invoke` tasks, i.e. functions with the `@task` decorator
+  - Put `# type: ignore` so that `mypy` does not complain
+
+- Example:
+  ```python
+  @task def run_qa_tests( # type: ignore 
+    ctx, 
+    stage="dev", 
+    version="", 
+  ):
```
samarth9008, line 65:
```python
@task 
def run_qa_tests( # type: ignore 
  ctx, 
  stage="dev", 
  version="", 
):
```

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+- We try to avoid it everywhere when possible
+
+## `np.array` and `np.ndarray`
+
+- If you get something like the following lint:
+  ```bash
+  dataflow/core/nodes/sklearn_models.py:537:[amp_mypy] error: Function "numpy.core.multiarray.array" is not valid as a type [valid-type]  
+  ```
+- Then the problem is probably that a parameter that the lint is related to
+  has been types as `np.array` while it should be typed as `np.ndarray`:
```
samarth9008, line 89: I think it should be **has been typed as**

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+  the following lint appears:
+  ```bash
+  dataflow/core/nodes/sklearn_models.py:155:[amp_mypy] error: "None" has no attribute "fit"
+  ```
+- A solution is to
+  - 1. Type hint when assigning the model parameter in ctor:
+     ```bash
+     self._model: Optional[sklearn.base.BaseEstimator] = None
+     ```
+  - 2. Cast a type to the model parameter after asserting that it is not `None`:
```
samarth9008, line 156: Is it okay to have number as well as bullet points? Wdyt @DanilYachmenev

DanilYachmenev: hmm, looked ok at first but now I think that there is no order to the points so lets remove the numbers

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+     ```bash
+     hdbg.dassert_is_not(self._model, None) self._model = cast(sklearn.base.BaseEstimator, self._model)
```
DanilYachmenev, line 158: also split on 2 lines
```
hdbg.dassert_is_not(self._model, None)
self._model = cast(sklearn.base.BaseEstimator, self._model)
```

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+  ```bash
+  self._model = self._model.fit(...)
+  ```
+  the following lint appears:
+  ```bash
+  dataflow/core/nodes/sklearn_models.py:155:[amp_mypy] error: "None" has no attribute "fit"
+  ```
+- A solution is to
+  - 1. Type hint when assigning the model parameter in ctor:
+     ```bash
```
DanilYachmenev, line 153: this one is not `bash` but `python`

`bash` is put for the cases when we put commands from bash calls or log outputs in cmd
all the code that is from the actual code base should be backticked with ```python

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+  ```
+
+# Handling the annoying `"None" has no attribute`
+
+- In some model classes `self._model` parameter is being assigned to `None` in
+  ctor and being set after calling `set_fit_state` method
+- The problem is that statically it's not possible to understand that someone
+  will call `set_fit_state` before using `self._model`, so when a model's method
+  is applied:
+  ```bash
```
DanilYachmenev, line 144: python

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+  dbg.dassert_is_not(var, None)
+  test_func(arg=var)
+  ```
+- Sometimes `mypy` doesn't pick up the `None` check, and warns that the function
+  expects a `bool` rather than an `Optional[bool]`. In that case, the solution
+  is to explicitly use `typing.cast` on the argument when passing it in, note
+  that `typing.cast` has no runtime effects and is purely for type checking.
+- Here're the relevant [docs](https://mypy.readthedocs.io/en/stable/casts.html)
+- So the solution would be:
+  ```bash
```
DanilYachmenev, line 131: python

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+
+## `np.array` and `np.ndarray`
+
+- If you get something like the following lint:
+  ```bash
+  dataflow/core/nodes/sklearn_models.py:537:[amp_mypy] error: Function "numpy.core.multiarray.array" is not valid as a type [valid-type]  
+  ```
+- Then the problem is probably that a parameter that the lint is related to
+  has been types as `np.array` while it should be typed as `np.ndarray`:
+  ```bash
```
DanilYachmenev, line 90: python

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+  `x_vals: np.array` -> `x_vals: np.ndarray`
+  ```
+
+# Handling the annoying `Incompatible types in assignment`
+
+- `mypy` assigns a single type to each variable for its entire scope
+- The problem is in common idioms where we use the same variable to store
+  different representations of the same data
+  ```bash
```
DanilYachmenev, line 99: python

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+  output : str = ...
+  output = output.split("\n")
+  ...
+  # Process output.
+  ...
+  output = "\n".join(output)
+  ```
+- Unfortunately the proper solution is to use different variables
+  ```bash
```
DanilYachmenev, line 108: python

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,231 @@
...
+  output : str = ...
+  output_as_array = output.split("\n")
+  ...
+  # Process output.
+  ...
+  output = "\n".join(output_as_array)
+  ```
+- Another case could be:
+  ```bash
```
DanilYachmenev, line 117: python

docs/Python_Type_hints.md
hhxjqm: There are numbers show in Gdoc, if it is not matter, I will remove them.

docs/Python_Type_hints.md
```diff
@@ -0,0 +1,233 @@
+# Python - Type hints
```
DanilYachmenev, line 1: Oh, and last thing
lets name it just `Type hints` and the doc name would be `Type_hints.md`

DanilYachmenev (commented):
change the name with filename and lgtm

samarth9008:
LGTM

hhxjqm:
@DanilYachmenev It's been two days, do I need to ask for new task?

samarth9008:
> @DanilYachmenev It's been two days, do I need to ask for new task?

We will file more and let you know. Now @gpsaggese will take a look and merge it.

gpsaggese (commented):
LG.

@DanilYachmenev and @samarth9008 feel free to strike out the gdocs / move them to OLD as I did for this, so we know what's left

PR #352, opened 2023-06-26
CmTask351_Improvement_to_the_encryption_flow

gpsaggese:


samarth9008:
@KangmingL 

Did you check? Is it working? As soon as you are done with that particular TODO remove it from the code.

KangmingL:
Got you, the encryption part works correctly. Still working on the tests.

samarth9008:
Also can you please explain how did you implement this TODO? `Make --test default.`

KangmingL:
@samarth9008 Check my latest commit, I remove TODOs, and make `--test` holds a default value `True`.

KangmingL:
@gpsaggese @samarth9008 For this 
"TODO(gp): Why is sudo needed? IMO the Docker container should use the right permissions."
I think `sudo` is needed, because `_tweak_init` is not running in a Docker container, so give Docker container the right permissions won't fix permission errors.

dev_scripts/encrypt_model.py
```diff
@@ -76,23 +83,36 @@ def _tweak_init(encrypted_model_dir: str) -> None:
...
+def _test_model(model_dir: str, model_dag_builder: str) -> None:
     """
     Check that model works correctly.
     """
     temp_file_path = "./tmp.encrypt_model.test_model.sh"
     import_path = model_dir.lstrip("./")
     import_path = import_path.replace("/", ".")
-    script = f'python -c "import {import_path}.mock1_pipeline as f; a = f.Mock1_DagBuilder(); print(a)"'
+    # TODO(gp): The model name should be passed from command line.
+    script = f'python -c "import {import_path}.C5a_pipeline as f; a = f.C5a_DagBuilder(); print(a)"'
```
samarth9008, line 115: I am not sure if I got this correctly but the general idea was to pass the `model_dag_builder` for testing but over here we are using the hardcoded name `C5a_pipeline`. Shouldn't we pass the `model_dag_builder`? Correct me if I am wrong.

KangmingL: Just ignore testing part for now, I'm still trying to figure this out. Also I realize the name of `.mock1_pipeline` may be different as well. Do you think I should add another args like `--model_pipline_name`?

samarth9008: Not needed. You specify the directory like this `dataflow_amp/pipelines/` and all the models inside it gets encrypted. Refer to that `dir` to see the better structure.

KangmingL: I mean GP said we cannot assume model name is `Mock1` or `C5a`, so the `model_dag_builder` should be something like `Mock1_DagBuilder` or `C5a_DagBuilder`. Therefore, in this part `import {import_path}.mock1_pipeline as f`, `mock1_pipeline` can be a different name. That's why I'm thinking passing in a `--model_pipeline_name`.

samarth9008: I think you are right but even I am not sure about this. Lets confirm it with @gpsaggese.

samarth9008: But I also wonder how will the testing part works? Bcz we are making it a default which will be performed all the time.

KangmingL: Actually I'm also concerned about that. It seems not very efficient. I guess just wait and see how GP responds to this testing part.

samarth9008:
Can be closed.

Issue #370, opened 2023-06-30
`invoke` tasks doesn't work from sorrentum


samarth9008:
None of the invoke tasks works from the sorrentum side after pulling the latest master.

Error:
```
(amp.client_venv) sorrentum@samarthkp:~/src/sorrentum1$ i docker_bash
One and only one set-up config should be true:
is_cmamp_prod=False
is_dev4=False
is_dev_ck=False
is_ig_prod=False
is_inside_ci=False
is_mac=False
INFO: > cmd='/home/sorrentum/src/venv/amp.client_venv/bin/invoke docker_bash'
## docker_bash: 
Traceback (most recent call last):
  File "/home/sorrentum/src/venv/amp.client_venv/bin/invoke", line 8, in <module>
    sys.exit(program.run())
  File "/home/sorrentum/src/venv/amp.client_venv/lib/python3.10/site-packages/invoke/program.py", line 398, in run
    self.execute()
  File "/home/sorrentum/src/venv/amp.client_venv/lib/python3.10/site-packages/invoke/program.py", line 583, in execute
    executor.execute(*self.tasks)
  File "/home/sorrentum/src/venv/amp.client_venv/lib/python3.10/site-packages/invoke/executor.py", line 140, in execute
    result = call.task(*args, **call.kwargs)
  File "/home/sorrentum/src/venv/amp.client_venv/lib/python3.10/site-packages/invoke/tasks.py", line 138, in __call__
    result = self.body(*args, **kwargs)
  File "/home/sorrentum/src/sorrentum1/helpers/lib_tasks_docker.py", line 1275, in docker_bash
    docker_cmd_ = _get_docker_compose_cmd(
  File "/home/sorrentum/src/sorrentum1/helpers/lib_tasks_docker.py", line 1121, in _get_docker_compose_cmd
    docker_cmd_ = _get_docker_base_cmd(
  File "/home/sorrentum/src/sorrentum1/helpers/lib_tasks_docker.py", line 1036, in _get_docker_base_cmd
    image = get_image(base_image, stage, version)
  File "/home/sorrentum/src/sorrentum1/helpers/lib_tasks_docker.py", line 930, in get_image
    base_image = _get_base_image(base_image)
  File "/home/sorrentum/src/sorrentum1/helpers/lib_tasks_docker.py", line 887, in _get_base_image
    _dassert_is_base_image_name_valid(base_image)
[...]

PR #371, opened 2023-06-30
SorrTask370 invoke tasks doesn't work from sorrentum

samarth9008:
#370 

- Currently pushed it according to the previous version of the code and its working. 
- Will work on regex and file a new PR


gpsaggese (commented):
Good move

Issue #355, opened 2023-06-26
Unit test `tee` param for `hsystem.system()`


DanilYachmenev:
Add unit tests for `system()` from `helpers/hsystem.py`

Before implementing read Unit tests [doc](https://github.com/sorrentum/sorrentum/blob/master/docs/Unit_tests.md) carefully

We need to add a test to `Test_system1` in `helpers/test/test_system_interaction.py` that runs `.system()` on a failing cmd with `--tee` switch enabled.

the idea is to test that if a failing command is executed with the `—tee` switch the cmd will fail (it had not been the case until we added the `PIPESTATUS` thing)
See how failing is being tested at `test7`

Assigning @samarth9008 to re-distribute
FYI @PomazkinG

Dylan-Celli:
Also can I include the tests where I pass a string, an int, and a float in the same method or should I put each in a different method?

DanilYachmenev:
> Also can I include the tests where I pass a string, an int, and a float in the same method or should I put each in a different method? 

Not sure what you mean, extrapolate

> Additionally could someone explain the naming convention behind the variable `act = str(cm.exception)` in `test7`? Just so I can be consistent with the naming convention.

`act` is for "actual" function outcome that should be compared with expected. Better name your actual as `actual`

DanilYachmenev:
> Is this cause for concern?

not sure personally
@samarth9008 do you know anything bout it?

samarth9008:
I am not sure. I think even I got it sometime.

Dylan-Celli:
> Not sure what you mean, extrapolate

I am creating a unit test to verify that if the `-tee` flag takes in a non-valid, non-boolean, value that the user gets an error thrown on the CLI. I want to create tests for incorrectly inputting strings, ints, and float values. I wanted to know if these tests should all have their own unit test method or if I can put them all into the same unit test method.

Dylan-Celli:
Hello I am running into trouble with the following command

`i run_fast_tests --coverage -p /home/dcelli/src/sorrentum1/helpers/test/test_system_interaction.py::Test_system1::test8`

When I run it I get the following error
```
ERROR: file or directory not found: /home/dcelli/src/sorrentum1/helpers/test/test_system_interaction.py::Test_system1::test8
ERROR: 4
```
But when I run the pytest command myself from the bash terminal the test is able to find the file and run normally

DanilYachmenev:
[...]

PR #378, opened 2023-07-02
Sorr task355 unit test tee param for hsystem system

Dylan-Celli:
@DanilYachmenev Here is the draft PR for the unit test I was asking about in [Sorr task355](https://github.com/sorrentum/sorrentum/issues/355) the code that I wrote is in `helpers/test/test_system_interaction.py::System_test1::test8`


DanilYachmenev:
@Dylan-Celli you need to put me to reviewers section, not assignees
since you're the assignee

DanilYachmenev:
remember about the labels as well

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,21 @@ def test7(self) -> None:
...
+        """        
+        with self.assertRaises(AssertionError) as cm:
+            hsystem.system("ls this_should_fail", tee=True)    
+        output = 'tmp_hsystem_Test_system1_test8_output.txt'
+        try:            
+            hsystem.system("ls this_should_fail", tee=True, output_file=output)
+        except Exception as e:
+            self.fail("hsystem.system() raised", type(e).__name__," error")
+        finally:            
+            os.remove(output)
```
DanilYachmenev, line 84: well, you should not call a tested command twice in one test
in this case, specify a cmd as a separate param first and check that error is thrown
then check its signature as it is done in `test7` above

no need for `try/except` clauses
no need to remove output as well
read what I wrote you about scratch dir in https://github.com/sorrentum/sorrentum/issues/355#issuecomment-1616780206

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,21 @@ def test7(self) -> None:
...
 
+    def test8(self) -> None:
+        """
+        Test .system() on a failing cmd with --tee switch enabled.
+        """        
+        with self.assertRaises(AssertionError) as cm:
+            hsystem.system("ls this_should_fail", tee=True)    
+        output = 'tmp_hsystem_Test_system1_test8_output.txt'
+        try:            
+            hsystem.system("ls this_should_fail", tee=True, output_file=output)
```
DanilYachmenev, line 80: basically this should go to another test method 
so in `test8` you check just `tee`
in `test9` you pass `tee` with a log file
and in `test10` you pass `tee` and `log_file` but also `allow_errors=True`

DanilYachmenev (commented):
see the comments, add other test methods

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,35 @@ def test7(self) -> None:
...
+        Test .system() on a failing cmd with --tee switch enabled and specified log output file.
+        """  
+        log_dir = self.get_scratch_space()
+        log_file_path = os.path.join(log_dir, "tee_log")   
+        #Creates AssertRaisesContext error (seems unexpected)
+        with self.assertRaises(RuntimeError) as cm:
+            hsystem.system("ls this_should_fail", tee=True, output_file=log_file_path)
+        # Check that log file contains error message.
+        log_content = hio.from_file(log_file_path)
+        self.check_string(log_content)
```
DanilYachmenev, line 92: Lets actually use the approach with signatures here
so do smth like
```
actual_log_content = hio.from_file(log_file_path)
expected_log_content = r"""
Log file text.
"""
self.assert_equal(actual_log_content, expected_log_content, fuzzy_match=True)
```

To figure out the log content you can just run the test with `-s --dbg` and see the expected log content in error logs
Just make sure logs make sense
```
> pytest helpers/test/test_system_interaction.py:: Test_system1:: test9 -s --dbg
```

same for `test10`

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,35 @@ def test7(self) -> None:
         act = re.sub(r"rc='(\d+)'", "rc=''", act)
         self.check_string(act)
 
+    def test8(self) -> None:
+        """
+        Test .system() on a failing cmd with --tee switch enabled and no other parameters.
```
DanilYachmenev, line 75:
```
"""
Check that an assert is raised when `tee` is passed without log file.
"""
```

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,35 @@ def test7(self) -> None:
...
+        """        
+        with self.assertRaises(AssertionError) as cm:
+            hsystem.system("ls this_should_fail", tee=True)    
+
```
DanilYachmenev, line 79: add exception signature check here as well
```
actual = str(cm.exception)
expected = r"""
Exception text.
"""
self.assert_equal(actual, expected, fuzzy_match=True)
```

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,35 @@ def test7(self) -> None:
...
+    
+    def test9(self) -> None:
+        """
+        Test .system() on a failing cmd with --tee switch enabled and specified log output file.
+        """  
+        log_dir = self.get_scratch_space()
+        log_file_path = os.path.join(log_dir, "tee_log")   
+        #Creates AssertRaisesContext error (seems unexpected)
+        with self.assertRaises(RuntimeError) as cm:
+            hsystem.system("ls this_should_fail", tee=True, output_file=log_file_path)
```
DanilYachmenev, line 89: do `rc = hsystem...`
and then assert its value
see https://github.com/sorrentum/sorrentum/issues/355#issuecomment-1616780206

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,35 @@ def test7(self) -> None:
...
+            hsystem.system("ls this_should_fail", tee=True)    
+        
+    
+    def test9(self) -> None:
+        """
+        Test .system() on a failing cmd with --tee switch enabled and specified log output file.
+        """  
+        log_dir = self.get_scratch_space()
+        log_file_path = os.path.join(log_dir, "tee_log")   
+        #Creates AssertRaisesContext error (seems unexpected)
```
DanilYachmenev, line 87: Replace with just `# Run.`

in general, keep in mind comments format
like space after `#`, avoid code references

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,35 @@ def test7(self) -> None:
...
+        """
+        Test .system() on a failing cmd with --tee switch enabled and no other parameters.
+        """        
+        with self.assertRaises(AssertionError) as cm:
+            hsystem.system("ls this_should_fail", tee=True)    
+        
+    
+    def test9(self) -> None:
+        """
+        Test .system() on a failing cmd with --tee switch enabled and specified log output file.
```
DanilYachmenev, line 83:
```
"""
Check that the failing command fails and logs are stored in the log file.

- `allow_errors = False`
- `tee = True`
- Log file path is passed
"""
```

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,35 @@ def test7(self) -> None:
...
+        # Check that log file contains error message.
+        log_content = hio.from_file(log_file_path)
+        self.check_string(log_content)
+
+    def test10(self) -> None:
+        """
+        Test .system() on a failing cmd with --tee switch and allowedErrors enabled.
+        """
+        with self.assertRaises(TypeError) as cm:
+            hsystem.system("ls this_should_fail", tee=True, allowedErrors=True)
```
DanilYachmenev, line 99: Well, this one should pass actaully
but see `system()` func interface - it does not have `allowedErrors` param
see  the func interface and https://github.com/sorrentum/sorrentum/issues/355#issuecomment-1616780206

Also do 
```
rc = hsystem.system(...
```
and then assert the value of `rc`

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,35 @@ def test7(self) -> None:
...
+        #Creates AssertRaisesContext error (seems unexpected)
+        with self.assertRaises(RuntimeError) as cm:
+            hsystem.system("ls this_should_fail", tee=True, output_file=log_file_path)
+        # Check that log file contains error message.
+        log_content = hio.from_file(log_file_path)
+        self.check_string(log_content)
+
+    def test10(self) -> None:
+        """
+        Test .system() on a failing cmd with --tee switch and allowedErrors enabled.
```
DanilYachmenev, line 96:
```
"""
Check that the failing command passes and logs are stored in the log file.

- `allow_errors = True`
- `tee = True`
- Log file path is passed
"""
```

Dylan-Celli:
@DanilYachmenev Here are the log signatures [you asked for](https://github.com/sorrentum/sorrentum/issues/355#issuecomment-1620278823)

Test 8 (successful)
```
INFO: > cmd='/venv/bin/pytest -s --dbg helpers/test/test_system_interaction.py::Test_system1::test8'

09:10:51 - INFO  hunit_test.py setUp:1042                               09:10:51 - INFO  hunit_test.py setUp:1042                               
################################################################################
Test_system1.test8
################################################################################
09:10:51 hwall_clock_time.py reset_current_bar_timestamp:94                      09:10:51 hwall_clock_time.py reset_current_bar_timestamp:94                      Reset
```

Test 9 (failed)
```
cmd='(ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}' failed with rc='2'

09:20:52 hwall_clock_time.py reset_current_bar_timestamp:94                      Reset
09:20:52 hsystem.py _system:224                              > (ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}
09:20:52 - ERROR hsystem.py _system:273  

FAILED helpers/test/test_system_interaction.py::Test_system1::test9 - AssertionError:
```

Test 10 (successful)
```
INFO: > cmd='/venv/bin/pytest -s --dbg helpers/test/test_system_interaction.py::Test_system1::test10'

################################################################################
Test_system1.test10
################################################################################
09:24:47 hwall_clock_time.py reset_current_bar_timestamp:94                      09:24:47 hwall_clock_time.py reset_current_bar_timestamp:94                      Reset

```

Dylan-Celli:
Also I was taking a look at `/helpers/hunit_test.py::Test_Case::check_string` and the code looks like it is supposed to take in a log file name and instead of the log file contents. Additionally, the code is throwing an error on line 1468 of `check_string` which is only available if the given file name does not exist. This does not make sense to me because we are passing an absolute path to the `check_string` method so the file should be found if the file path is passed in with the variable `log_file_path`

DanilYachmenev:
@Dylan-Celli these are not log signatures but logs themselves

Signature is a string representation of an object that can be compared in the test after
E.g. see how signatures are used in `helpers/test/test_hpandas.py::Test_dassert_is_unique1`

Anyway, there are still comments for you to implement, so lets do them first and them we will be able to take a closer look
now try to follow them and change label only when you want me to review the whole PR

DanilYachmenev:
> Also I was taking a look at `/helpers/hunit_test.py::Test_Case::check_string` and the code looks like it is supposed to take in a log file name and instead of the log file contents. Additionally, the code is throwing an error on line 1468 of `check_string` which is only available if the given file name does not exist. This does not make sense to me because we are passing an absolute path to the `check_string` method so the file should be found if the file path is passed in with the variable `log_file_path`

No need to use `check_string()`, see https://github.com/sorrentum/sorrentum/pull/378#discussion_r1252032426
We want to use signature comparison approach now

helpers/test/test_system_interaction.py
Dylan-Celli: In the `test9` unit test errors are not permitted and therefore the command should raise an error, correct? If so I am confused as to why I am trying to get the returned value `rc` from the failed `hsystem.system()` method because this method should not return anything due to the exception raised within it. I only ask because I want to make sure I am understanding the scenario correctly.

DanilYachmenev: yep, you're right, in this case `rc` should not be set, since cmd fails
so no reason to do `rc = ` here
it is needed in the next test though

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,64 @@ def test7(self) -> None:
...
+            rc = hsystem.system(
+                "ls this_should_fail", tee=True, output_file=log_file_path
+            )
+        actual = str(cm.exception)
+        expected = r"""
+        cmd='(ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}' failed with rc='2'
+        truncated output=
+        ls: cannot access 'this_should_fail': No such file or directory
+        """
+        self.assert_equal(actual, expected, fuzzy_match=True)
```
DanilYachmenev, line 109: And also after this check we need to read `log_file_path` and check its signature in the same way you did for the exception.
Despite the fact that func has failed, it should've written the error log in the log file.
My expectation would be that it is the same as `expected` content here, although it might be not true.

In any case, check it by extracting `log_file_path` content and checking its signature

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,64 @@ def test7(self) -> None:
...
+        """
+        log_dir = self.get_scratch_space()
+        log_file_path = os.path.join(log_dir, "tee_log")
+        rc = hsystem.system(
+            "ls this_should_fail",
+            tee=True,
+            abort_on_error=False,
+            output_file=log_file_path,
+        )
+        self.assertTrue(rc == 2)
```
DanilYachmenev, line 128: same here
read `log_file_path` and check its signature

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,64 @@ def test7(self) -> None:
...
+        Check that the failing command fails and logs are stored in the log
+        file.
+
+        - `allow_errors = False`
+        - `tee = True`
+        - Log file path is passed
+        """
+        log_dir = self.get_scratch_space()
+        log_file_path = os.path.join(log_dir, "tee_log")
+        # Run
```
DanilYachmenev, line 98: Remove

and overall - finish all comments with `.`

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,64 @@ def test7(self) -> None:
...
+        Check that an assert error is raised when `tee` is passed without a log file.
+        """
+        with self.assertRaises(AssertionError) as cm:
+            rc = hsystem.system("ls this_should_fail", tee=True)
+        actual = str(cm.exception)
+        expected = r"""
+        ################################################################################
+        * Failed assertion *
+        'True' implies 'False'
+        ################################################################################
```
DanilYachmenev, line 83: Quite uninformative log tbh
it passes this way, right?

DanilYachmenev (commented):
looking good
see the comments below - we need to check log file content
and done

helpers/test/test_system_interaction.py
Dylan-Celli: Got it! I will keep that in mind going forward

helpers/test/test_system_interaction.py
Dylan-Celli: Does it make sense if the CLI logs are slightly different from the file logs?
CLI log:
```
cmd='(ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}' failed with rc='2'
        truncated output=
        ls: cannot access 'this_should_fail': No such file or directory
```
file output log:
```
ls: cannot access 'this_should_fail': No such file or directory\n
```

helpers/test/test_system_interaction.py
Dylan-Celli: I think the reason why it is so vague is because of line 177 of `hsystem.py`
```
hdbg.dassert_imply(tee, output_file is not None)
```
where output file for this test is actually `None`

helpers/test/test_system_interaction.py
DanilYachmenev: yes, seems ok

helpers/test/test_system_interaction.py
gpsaggese, line 128: -> assertEqual(rc, 2)

so you get a better error message

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,74 @@ def test7(self) -> None:
...
+        expected = r"""
+        cmd='(ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}' failed with rc='2'
+        truncated output=
+        ls: cannot access 'this_should_fail': No such file or directory
+        """                
+        self.assert_equal(actual, expected, fuzzy_match=True)  
+        # Getting log output from output file.
+        expected = r"ls: cannot access 'this_should_fail': No such file or directory\n"          
+        with open(log_file_path,'r') as log_file_content:
+            actual = log_file_content.read()
```
DanilYachmenev, line 112: instead do `actual = hio.from_file(log_file_path)`
here and below

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,74 @@ def test7(self) -> None:
...
+            )
+        actual = str(cm.exception)
+        expected = r"""
+        cmd='(ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}' failed with rc='2'
+        truncated output=
+        ls: cannot access 'this_should_fail': No such file or directory
+        """                
+        self.assert_equal(actual, expected, fuzzy_match=True)  
+        # Getting log output from output file.
+        expected = r"ls: cannot access 'this_should_fail': No such file or directory\n"
```
DanilYachmenev, line 110: put line with `expected` under `actual` for consistency

here and below

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,74 @@ def test7(self) -> None:
...
+                "ls this_should_fail", tee=True, output_file=log_file_path
+            )
+        actual = str(cm.exception)
+        expected = r"""
+        cmd='(ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}' failed with rc='2'
+        truncated output=
+        ls: cannot access 'this_should_fail': No such file or directory
+        """                
+        self.assert_equal(actual, expected, fuzzy_match=True)  
+        # Getting log output from output file.
```
DanilYachmenev, line 109: -> `# Check log output.`

here and below

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +69,74 @@ def test7(self) -> None:
...
+        """
+        log_dir = self.get_scratch_space()
+        log_file_path = os.path.join(log_dir, "tee_log")
+        rc = hsystem.system(
+            "ls this_should_fail",
+            tee=True,
+            abort_on_error=False,
+            output_file=log_file_path,
+        )        
+        self.assertEqual(rc,2)
```
DanilYachmenev, line 133: run linter to fix spacings
a minor nit, but params in functions are also separated with space after comma
so `...(rc, 2)`
linter should do it automatically

DanilYachmenev (commented):
almost there

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,74 @@ def test7(self) -> None:
...
+            )
+        actual = str(cm.exception)
+        expected = r"""
+        cmd='(ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}' failed with rc='2'
+        truncated output=
+        ls: cannot access 'this_should_fail': No such file or directory
+        """
+        self.assert_equal(actual, expected, fuzzy_match=True)
+        # Check log output.
+        with open(log_file_path, "r") as log_file_content:
```
DanilYachmenev, line 112: no need for this line rn
just
```
# Check log output.
actual = hio.from_file(log_file_path)
expected = r"ls: cannot access 'this_should_fail': No such file or directory\n"
self.assert_equal(actual, expected, fuzzy_match=True)
```

same below

DanilYachmenev (commented):
lgtm

@PomazkinG wdyt? are these test sufficient?

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,72 @@ def test7(self) -> None:
...
+        """
+        log_dir = self.get_scratch_space()
+        log_file_path = os.path.join(log_dir, "tee_log")
+        with self.assertRaises(RuntimeError) as cm:
+            rc = hsystem.system(
+                "ls this_should_fail", tee=True, output_file=log_file_path
+            )
+        actual = str(cm.exception)
+        expected = r"""
+        cmd='(ls this_should_fail) 2>&1 | tee -a /app/helpers/test/outcomes/Test_system1.test9/tmp.scratch/tee_log; exit ${PIPESTATUS[0]}' failed with rc='2'
```
gpsaggese, line 106: @DanilYachmenev and @PomazkinG this should be ok when purified while running in Orange. If not it's an easy fix. Let's just merge

helpers/test/test_system_interaction.py
```diff
@@ -69,6 +70,72 @@ def test7(self) -> None:
...
+        """
+        log_dir = self.get_scratch_space()
+        log_file_path = os.path.join(log_dir, "tee_log")
+        rc = hsystem.system(
+            "ls this_should_fail",
+            tee=True,
+            abort_on_error=False,
+            output_file=log_file_path,
+        )
+        self.assertEqual(rc, 2)
```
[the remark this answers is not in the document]
gpsaggese: Agree

gpsaggese (commented):
Minor tweaks and merge for me. Good work

helpers/test/test_system_interaction.py
PomazkinG: agree

@DanilYachmenev try `purify_txt_from_client()`

DanilYachmenev: @Dylan-Celli pls do 
```
actual = str(cm.exception)
actual = hunitest.purify_txt_from_client(actual)
expected = r"""
...
```
and update `expected` accordingly (rerun the test)
when done, I'll do the check at `orange`

Dylan-Celli: When I use the `i docker_bash` the docker container does not setup with my username, it did not affect my testing before. So when I log in the prompt says `I have no name!@263ec60fb41a:/app$`  which causes an error with the `hunitest.purify_txt_from_client()` since the container does not actually have my username. I went through the setup documentation for to see if I was missing anything and setup my environment again but I am still running into this issue. I looked through the setup log for the `i docker_bash` and found 
```
One and only one set-up config should be true:
is_cmamp_prod=False
is_dev4=False
is_dev_ck=False
is_ig_prod=False
is_inside_ci=False
is_mac=False
Don't recognize host: host_os_name=Linux, am_host_os_name=Linux
```
Is there something that I am missing? My sorrentum folder is in `/home/dcelli/src/sorrentum1`.

DanilYachmenev: @Dylan-Celli not sure bout reason for your problem
did you set up your env?
`> source dev_scripts/setenv_amp.sh`?

In any case, I've implemented the change myself so take a look

DanilYachmenev (commented):
@Dylan-Celli further on do not forget to run linter and fix the lints before every commit

PR #401, opened 2023-07-07
Cm task283 create test list to run with sorrentum

alejandroBallesterosC:
added pytest markers for tests that require docker to run. Rebased changes from master since last commit.


gpsaggese:
This is busted closing.

PR #463, opened 2023-07-26
Fix few VM issues within Quick_start_for_developing.md

aaryapatel100:


DanilYachmenev:
@aaryapatel100 what issue is this related to btw?
I expect this is a follow-up from the onboarding, right?

if so, pls put a reference to it in the PR desc
also pls read https://github.com/sorrentum/sorrentum/blob/master/docs/First_review_process.md to see how to treat the PR to make sure that it is not unintentionally overlooked

docs/Quick_start_for_developing.md
```diff
@@ -112,7 +115,9 @@
 
   - Linux sudo problem, see
     [here](https://stackoverflow.com/questions/48568172/docker-sock-permission-denied)
-    for the solution
+    for the solution (or
+    [here](https://phoenixnap.com/kb/docker-permission-denied.)
+    for issues within the thin environment)
```
samarth9008, line 120: I am not sure how the link you provided is related to the issue within the thin environment because it is still related to the docker issue.

docs/Quick_start_for_developing.md
```diff
@@ -121,19 +126,29 @@
...
+  2. Build the thin environment; this is done once
 
-  2. Activate the thin environment; make sure it is always activated
+     ```
+     > source dev_scripts/client_setup/build.sh
+     ```
+     - If terminal crashes from
+       [aws-cli installation error](https://github.com/aws/aws-cli/issues/8036)
+       with output ending in `AttributeError: cython_sources`, open dev_scripts/client_setup/requirements.txt
+       and insert `pyyaml==5.3.1` at the top
```
samarth9008, line 143: If terminal crashes from [aws-cli installation error](https://github.com/aws/aws-cli/issues/8036)
       with output ending in `AttributeError: cython_sources`, insert  `pyyaml==5.3.1` in `dev_scripts/client_setup/requirements.txt`

docs/Quick_start_for_developing.md
```diff
@@ -208,7 +223,13 @@
   - pull the latest container with `invoke docker_pull`
   ```
 
-  2. Start a Jupyter server
+  2. Exit out of app terminal prompts
+
+     ```
+     I have no name!@60a1f38f89a3:/app$ exit
+     ```
```
samarth9008, line 230: I do not think this is needed bcz in some doc it is mentioned that any command starting with `>` should be run on the local terminal and `docker>` should run inside docker container.

FYI, whenever you see terminal like this `I have no name!@60a1f38f89a3:/app$`, it means you are inside docker container.

samarth9008 (commented):
Although, I have reviewed but lets wait for @DanilYachmenev comment before implementing any changes.

docs/Quick_start_for_developing.md
aaryapatel100: Ok, sorry should've clarified as just an additional resource. I added that link because the steps in the other didn't seem to resolve the issue at the time.

docs/Quick_start_for_developing.md
aaryapatel100: Ah ok, thank you for the clarification.

DanilYachmenev:
> Although, I have reviewed but lets wait for @DanilYachmenev comment before implementing any changes.

honestly, I'm a bit out of context here
@aaryapatel100 could you extrapolate, what this changes are related to at all?

aaryapatel100:
Yeah, so these were all issues/errors that I encountered in the onboarding process in which I thought the steps might be better clarified in the documentation. Important to note, that these did come up within VMWare usage so not sure if errors will replicate for all but in the case that they do easy things to look out for. The aws-cli installation error during thin environment build is probably the most notable one as it just crashed the terminal in less than a second of outputting error and seems to be a much larger issue with this specific package dependency. I figured that adding a note about terminal crashing would be useful for 2 reasons: a) saves time for onboarder from having to screen record build process and see what happened and b) given the speed of crash, someone might figure it's not necessarily an error, open terminal, run the activation command, and deal with a whole host of other issues.

DanilYachmenev:
ok then implement the comments that Samarth has left and we can ask for GP to take a look

aaryapatel100:
Ok, comments have been implemented. @gpsaggese Could you please take a look to the updated quick-start documentation?

samarth9008:
Closing as not planned.

PR #464, opened 2023-07-27
Handle root path purification outliers

critrange:


DanilYachmenev:
@critrange pls go over this [doc](https://github.com/sorrentum/sorrentum/blob/master/docs/First_review_process.md) to see how to format your PR (with label, issue reference etc)

helpers/hunit_test.py
```diff
@@ -418,13 +418,13 @@ def purify_from_environment(txt: str) -> str:
         # Replace the git path with `$GIT_ROOT`.
         super_module_path = hgit.get_client_root(super_module=super_module)
         if super_module_path != "/":
-            txt = txt.replace(super_module_path, "$GIT_ROOT")
+            txt = txt.replace(super_module_path + "/", "$GIT_ROOT" + "/")
```
DanilYachmenev, line 421: cunningly done)
However, I'm afraid that it might not be robust enough
e.g. what if all we have in the output is a supermodule path that is not followed by `/`?
just `dir = /app`, for example

IMO we need to use a regex that replaces `super_module_path` string appearances in the text if it is NOT followed by a letter or number (conceding the punctuation like `.`, `,` or `\s` are ok)
@PomazkinG @gpsaggese wdyt?

helpers/hunit_test.py
```diff
@@ -418,13 +418,13 @@ def purify_from_environment(txt: str) -> str:
...
         else:
             # If the git path is `/` then we don't need to do anything.
             pass
     # Replace the current path with `$PWD`
     pwd = os.getcwd()
-    txt = txt.replace(pwd, "$PWD")
+    txt = txt.replace(pwd + "/", "$PWD" + "/")
```
DanilYachmenev, line 427: same here

helpers/test/test_unit_test.py
```diff
@@ -1181,3 +1181,20 @@ def test1(self) -> None:
             of class '_Man' is not a subclass of '<class 'int'>'
         """
         self.helper(txt, exp)
+
+
+class Test_purify_from_environment(hunitest.TestCase):
```
DanilYachmenev, line 1186: -> `Test_purify_from_environment1`
we end uni test funcs and methods with digits
to ease the processing

helpers/test/test_unit_test.py
```diff
@@ -1181,3 +1181,20 @@ def test1(self) -> None:
...
+    def helper(self, txt: str, exp: str) -> None:
+        act = hunitest.purify_from_environment(txt)
+        self.assert_equal(act, exp)
+
+    def test1(self) -> None:
+        """
+        Check wrong replace the git path with `$GIT_ROOT`.
```
DanilYachmenev, line 1193: IMO here it is enough 
```
"""
Check that the text is purified correctly.
"""
```

helpers/test/test_unit_test.py
```diff
@@ -1181,3 +1181,20 @@ def test1(self) -> None:
...
+
+        eg. jupyter_core/application.py maybe replaced with
+        jupyter_core$GIT_ROOTlication.py
+        """
+        txt = "jupyter_core/application.py"
```
DanilYachmenev, line 1198: lets use an example that reflects more
like
```
txt = "/app/jupyter_core/application.py"
exp = "$GIT_ROOT/jupyter_core/application.py"
```

helpers/test/test_unit_test.py
```diff
@@ -1181,3 +1181,20 @@ def test1(self) -> None:
             of class '_Man' is not a subclass of '<class 'int'>'
         """
         self.helper(txt, exp)
+
+
+class Test_purify_from_environment(hunitest.TestCase):
+    def helper(self, txt: str, exp: str) -> None:
+        act = hunitest.purify_from_environment(txt)
+        self.assert_equal(act, exp)
```
DanilYachmenev, line 1189: imo no need for `helper` here
just implement in the test directly

helpers/hunit_test.py
DanilYachmenev, line 427: add a test for this case as well

DanilYachmenev (commented):
smart move!
but lets take time to try implementing a more robust approach
although, it's just my paranoia, so called the guys to share thoughts

meanwhile see the comments

helpers/test/test_unit_test.py
gpsaggese: Ok to leave. Let's just call it with a better name, say "check_output"

critrange:
@DanilYachmenev So I should make new PR after updating my code. right?

helpers/hunit_test.py
PomazkinG: agree, e.g.,

`/app` -> `$GIT_ROOT`
`/application.py` -> `/application.py`

the regex logic could be: replace a path to the git root if:
- it is followed by `/`, e.g., `/app/test.txt`
- if it is a single word, e.g., `/app`

I would add also add a unit test for the `/app` case then

DanilYachmenev:
> @DanilYachmenev So I should make new PR after updating my code. right?

No need in a new PR
Just check out to your current branch, commit n push your changes
then you will see that most comments here will get a plank `Outdated` on top of them
And then you just press `Resolve conversation` on the each resolved one

helpers/hunit_test.py
critrange:
```python
def purify_from_environment(txt: str) -> str:
    # We remove references to the Git modules starting from the innermost one.
    for super_module in [False, True]:
        # Replace the git path with `$GIT_ROOT`.
        super_module_path = hgit.get_client_root(super_module=super_module)
        if super_module_path != "/":
            txt = txt.replace(super_module_path, "$GIT_ROOT")
        else:
            # If the git path is `/` then we don't need to do anything.
            pass
    # Replace the current path with `$PWD`
    pwd = os.getcwd()
    txt = txt.replace(pwd, "$PWD")
...
```
When run pytest for `purify_from_environment` inside docker container, $PWD can't be replaced because both `super_module_path` and `pwd` are "/app".
In other word, all "/app" are replaced with "$GIT_ROOT" and then nothing to be replaced with "$PWD".
So I am not sure how to add test for "$PWD" replacement.

PomazkinG: ok to test only `$GIT_ROOT`

Ideally we should split `purify_from_environment()` into 3 functions:
1. One replaces `$GIT_ROOT`
2. One replaces `$PWD`
3. One replaces `$USER_NAME`

helpers/hunit_test.py
```diff
@@ -418,13 +418,15 @@ def purify_from_environment(txt: str) -> str:
         # Replace the git path with `$GIT_ROOT`.
         super_module_path = hgit.get_client_root(super_module=super_module)
         if super_module_path != "/":
-            txt = txt.replace(super_module_path, "$GIT_ROOT")
+            pattern = re.compile(super_module_path + r"(?![\w])")
```
PomazkinG, line 421: nit: can we use f-string? since the regex is used multiple times I would factor it out. 

E.g.,
```
# Add a comment briefly explaining what the regex does.
dir_pattern = r"(?![\w])"
```

and below you do smth like:
```
# Same for pwd.
re.compile(f"{super_module_path}{dir_pattern}")
```

helpers/test/test_unit_test.py
```diff
@@ -1181,3 +1181,17 @@ def test1(self) -> None:
...
+class Test_purify_from_environment1(hunitest.TestCase):
+    def check_output(self, txt: str, exp: str) -> None:
+        act = hunitest.purify_from_environment(txt)
+        self.assert_equal(act, exp)
+
+    def test1(self) -> None:
+        """
+        Check that the text is purified correctly.
+        """
+        txt = "/app, /app/jupyter_core/application.py"
```
PomazkinG, line 1195: nit: I would just create `test2` for `/app` purely and keep only "/app/jupyter_core/application.py" here

PomazkinG (commented):
Nits only and it's done

@DanilYachmenev can you please run the regressions from the dev server to make sure we don't break anything?

DanilYachmenev:
> Nits only and it's done
> 
> @DanilYachmenev can you please run the regressions from the dev server to make sure we don't break anything?

run all tests from dev on `sorrentum`
everything passes (except for a couple unrelated tests), can be merged IMO

PR #350, opened 2023-06-26
SorrTask258_Replace_pytest_raises_with_self_assertraises_new

KangmingL:

PR #473, opened 2023-07-28
Sorr task283 create test list to run with sorrentum

LibertasSpZ:
Some progress in issue #283 

 - Fix `core/plotting/test/test_plots.py` by making it call the correct function and updating deprecated aliases along the way.
   - I ran `pytest core/plotting/test/test_plot.py` in `Sorrentum` and passed all 4 tests. 
   - Had to `pip install` a tone of modules.
 - Found many other failures are `File '/home/.aws/credentials' doesn't exist`. 
   - Will add `@pytest.mark.need_credentials` to those tests later.
 - In general I think it's going to take a huge amount of time to fix all the remaining 98 failures. The `test_plots.py` alone took about 2 hours to fix.


DanilYachmenev (commented):
well, actually, lets fix the test here
just remove the todos after you do the necessary checks

LibertasSpZ:
Done removing the todos. @DanilYachmenev

core/plotting/test/test_plots.py
```diff
@@ -71,13 +71,13 @@ def test_plot_time_series_by_period2(self) -> None:
         test_series = self.get_plot_time_series_by_period1()
         period = "time"
         cplmiplo.plot_time_series_by_period(test_series, period)
-
+
```
DanilYachmenev, line 74: oh, and also not sure if this change is needed

core/signal_processing/special_functions.py
```diff
@@ -13,8 +13,7 @@
 
 _LOG = logging.getLogger(__name__)
 
-
-def calculate_tau_from_com(com: float) -> Union[float, np.float]:
+def calculate_tau_from_com(com: float) -> Union[float, np.float64]:
```
DanilYachmenev, line 16: did you run linter on this file, right?

DanilYachmenev (commented):
just couple formatting nits
otherwise lgtm

gpsaggese (commented):
Can you report the error from the linter / pytest?

If np.float is being deprecated we should do a search / replace in the entire code base

LibertasSpZ:
The linter error was a false alarm: I didn't notice that my docker API was closed. Once I reopen the docker API there was no error on the linter's part.

pytest says np.float is deprecated, and after changing it to np.float64 it fixed at least one failing test. I can try searching and replacing later if you want.

gpsaggese:
> pytest says np.float is deprecated, and after changing it to np.float64 it fixed at least one failing test. I can try searching and replacing later if you want.

Ok, but can you post what exactly `pytest` is reporting? If it's just a warning, we will do a global replacement just than just this

LibertasSpZ:
@gpsaggese Sure. After I change Line 17 of `core/signal_processing/special_functions.py` back to `np.float`, I run `pytest core/plotting/test/test_plots.py` [which eventually calls `special_functions.py::calculate_tau_from_com` on line 17 of `special_functions.py`], and get the attached 
[np.float.txt](https://github.com/sorrentum/sorrentum/files/12212565/np.float.txt)
 results.

DanilYachmenev:
@LibertasSpZ could you pls make short replacement script so we could make this update in all our repos easily?
For this you have to create an `.sh` script that uses `dev_scripts/replace_text.py`
Pls run it on your end as well to see if it does all the replacements correctly
we need to replace only type hints and avoid changing the actual function calls of `np.float()`

Example PR
https://github.com/sorrentum/sorrentum/pull/350

LibertasSpZ:
Thanks for the task, @DanilYachmenev . Here are some progress:

[1] I ran `grep -r np.float .` in the ambient directory sorrentum1 and found that all np.float have been converted to np.float64 already by someone. See the attached output 
[np_float_type_report.txt](https://github.com/sorrentum/sorrentum/files/12225866/np_float_type_report.txt).  So there should probably be no need for further replacement.

[2] I tried to make a `.sh` file for replacing `np.float` to `np.float64` anyway and found an issue: currently `replace_text.py` is inside `sorrentum1/dev_scripts`, and `replace_text.py` imports `sorrentum1/helpers/`, which is NOT inside `sorrentum1/dev_scripts` anymore. Without either moving the `helpers/` back to `sorrentum1/dev_scripts` or modifying `replace_text.py` to manipulate `sys.path` in python, I don't think the `.sh` file can run correctly.

My `.sh` file is as attached 
[SorrTask283_Create_test_list_to_run_with_Sorrentum.txt](https://github.com/sorrentum/sorrentum/files/12225935/SorrTask283_Create_test_list_to_run_with_Sorrentum.txt). I'm changing the extension to `.txt` just to allow upload. It was placed in `sorrentum1/dev_scripts/cleanup_scripts`, and got the following error msg:

    ```
    + python3 ../replace_text.py --old 'np.float]' --new 'np.float64]' --ext _all_
    File "/Users/Libertas/src/sorrentum1/dev_scripts/cleanup_scripts/../replace_text.py", line 51, 
    in <module>
    import helpers.hdbg as hdbg
    ModuleNotFoundError: No module named 'helpers'
    ```
    As mentioned above, this indicates that `replace_text.py` cannot find `../helpers`, and looks like the only cure is to manipulate the code using `sys.path` if we don't move the directories around.

DanilYachmenev:
@LibertasSpZ 

[0] general comment: lets file PRs with implemented actions with the code changes and post log outputs in comments there
- it is much more convenient to review and give comments on the code in PR interface instead of txt
- no need to make a file with full logs - only most important lines of it are enough

txt files are hard to process and even use ctrl+F there to find some words
so lets try to avoid them whenever possible and just use GH interface for convenience

[1] yeah, I believe you've replaced all the instances in your PR. However, no we need to make a script that would do the same job you did manually.
Since in other repos there are the same changes needed + there are other files where the same pattern might apear and needs to be replaced.
Therefore, pls prepare the script in `.sh` and commit in a PR

[2] in this case everything should be OK. This is our general approach to imports and `/helpers` should not be inside `/dev_scripts` to be imported by a file located inside `/dev_scripts`
E.g. `import helpers.hdbg as hdbg` is importing the `/helpers/hdbg`  to the file and can be used without problems
you can read more about em here https://github.com/sorrentum/sorrentum/blob/master/docs/Imports_and_packages.md 

[3] I see your script and cannot comment it directly yet
but it is definitely a bit different from how it used to be done
check out `dev_scripts/cleanup_scripts/SorrTask258_Replace_pytest_raises_with_self_assertraises.sh` and ismilar files in the `dev_scripts/cleanup_scripts/` dir to see the format and do +- the same (like without `python3` etc)

so yep, please, do a PR with the script, consider the comments above, use ref code and share error logs if it aint gonna work

LibertasSpZ:
Thanks for the comments, @DanilYachmenev . I changed the `.sh` file to the desired format and pushed it. Still not working, but for a different reason.

As the error output attached here 
[errors.txt](https://github.com/sorrentum/sorrentum/files/12233188/errors.txt)
shows, now it runs the python, finds the .py files [to avoid massive undesired changes I am only testing inside the current folder]; it is supposed to make changes to the `testnpfloat.py` file I wrote, which I also pushed, but it failed to do any changes.

My conjecture is that the regex parsing is wrong. I tried using something like `np.float(?!6|\()` for `--new`, which is supposed to be the regex for "np.float not immediately followed by 6 and not immediately followed by (", [because we don't want to change `np.float64` to `np.float6464`, or `np.float(arg)` to `np.float64(arg)`], didn't work either.

Thanks.

dev_scripts/cleanup_scripts/SorrTask283_Create_test_list_to_run_with_Sorrentum.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -xe
+
+
+replace_text.py \
+    --old "np.float]" \
+    --new "np.float64]" \
+    --only_dirs "." \
```
DanilYachmenev, line 7: I suppose that the script may not work due to this line
try to do
```
--only_dirs "dev_scripts"
```
or just remove this one
the most important part here would be to set a correct regex

I suppose it should be `np\.float(?!6|\()` but pls double check

DanilYachmenev (commented):
good job with the test file
see the comment below

gpsaggese:
Guys, everything in this thread is correct, but I made a point earlier that I didn't understand why this is triggering only on @LibertasSpZ laptop. In fact from his log, he is running outside Docker (that's why he said "there are a lot of missing modules").
```
platform darwin -- Python 3.11.4, pytest-7.4.0, pluggy-1.2.0 -- /Users/Libertas/src/venv/amp.client_venv/bin/python3
```
All the development happens inside Docker which contains all the dependencies pinned down.
To run tests you need to do
```
> i docker_bash
docker> pytest
```
or directly `i run_fast_tests`

Let's merge this as it is, but it's a speculative fix for when we upgrade to whichever version of numpy deprecated np.float

Ok?

LibertasSpZ:
Thanks for catching this, @gpsaggese . I resolved the conversation for you to merge.

Will switch to [the other 283 branch](https://github.com/sorrentum/sorrentum/tree/SorrTask283_Create_test_list_to_run_with_Sorrentum_2) and try to get at least some regression passing. Will also make sure to do the development inside a docker container in future.

samarth9008:
@DanilYachmenev Can this closed?

DanilYachmenev:
> @DanilYachmenev Can this be closed?

@gpsaggese wdyt? do we still need this one?

samarth9008:
Closing this for now. Will re-open if needed.

Issue #466, opened 2023-07-27
Test dassert_file_extension()


DanilYachmenev:
The function is: https://github.com/sorrentum/sorrentum/blob/master/helpers/hdbg.py#L794C5-L794C27

Do not forget to remove the TODO

Create a new test class in https://github.com/sorrentum/sorrentum/blob/master/helpers/test/test_dbg.py

Test cases:
- [x] Extension is string: e.g., `.csv.gz` and file name ends with `.csv.gz`
- [x] Extension is a list: e.g., `[“.csv.gz”, “csv”]` and file name ends with `csv`
- [x] Incorrect extension: e.g., file name ends with `.txt` while the extensions are “.csv, .csv.gz”
    - Use self.AssertRaises to catch an assertion, e.g., see https://github.com/sorrentum/sorrentum/blob/master/helpers/test/test_dbg.py#L63

Read more about unit tests in the corresponding [doc](https://github.com/sorrentum/sorrentum/blob/master/docs/Unit_tests.md)

FYI @PomazkinG

dreamcatcher0101:
For the Case1:
If we assume that the filename ends with `.csv.gz`, the extension must be `.gz`.
Then the result is `failed`. I think it's needed to modify the `dassert_file_extension()` from [helpers/test_dbg.py](https://github.com/sorrentum/sorrentum/blob/master/helpers/test/test_dbg.py). Is it Ok to modify?

PomazkinG:
@dreamcatcher0101 feel free to modify

the expected outcome is:
- file name ends with `.csv.gz`
- `.csv.gz` is in the list of extensions
- no assertion is raised

DanilYachmenev:
> Hi @DanilYachmenev For the Case1: If we assume that the extension is `.csv.gz` and the filename ends with `.csv.gz`, the extension must be `.gz`. Then the result is `failed`. Is it correct? I think it's needed to modify the `dassert_file_extension()` from [helpers/test_dbg.py](https://github.com/sorrentum/sorrentum/blob/master/helpers/test/test_dbg.py) to fix it. Is it Ok to modify?

Yes, `act_ext = os.path.splitext(file_name)[-1].lower()` in the function is not processing `.csv.gz` extensions correctly
so modify it and add a unit test for such case pls

DanilYachmenev:
done via #477, closing

PR #465, opened 2023-07-27
Refactored purify_from_environment function to handle user name purification outliers

dreamcatcher0101:
### This PR resolve the [issue446](https://github.com/sorrentum/sorrentum/issues/446)

**How it was resolved?**
- Modified purify_from_environment() from [helpers/hunit_test.py](https://github.com/sorrentum/sorrentum/pull/465/files#diff-de839a7d62b39eeae08f3cb65e3dca7781216e5548c27201ca60f7f98d312fdf) by using splitter.
- Added [test cases](https://github.com/sorrentum/sorrentum/pull/465/files#diff-42f211e18d410108a66626483fb3678b670a3e11e1d2058a8b6fbf46eb473d2c) for modified function.

PR #477, opened 2023-07-28
Added test cases for dassert_file_extension()

dreamcatcher0101:
### This PR resolved the issue [#466](https://github.com/sorrentum/sorrentum/issues/466) 

**How it was resolved?**
- Added case to check if filename has given extension.
> Modified dassert_file_extension() to get correct extension
- Added case to check if filename has one of given extensions.
- Added case to check if assertion is raised when filename doesn't have one of given extensions.

### DIFF
--- a/helpers/hdbg.py
+++ b/helpers/hdbg.py
@@ -806,7 +806,7 @@ def dassert_file_extension(
     # Make sure extension starts with .
     extensions = [f".{e}" if not e.startswith(".") else e for e in extensions]
     # Check.
-    act_ext = os.path.splitext(file_name)[-1].lower()
+    act_ext = file_name[file_name.find("."):]
     dassert_in(
         act_ext,
         extensions,
--- a/helpers/test/test_dbg.py
+++ b/helpers/test/test_dbg.py
@@ -624,3 +624,42 @@ class Test_dassert_related_params2(hunitest.TestCase):
         message 'hello world'
         """
         self.assert_equal(act, exp, purify_text=True, fuzzy_match=True)
+
+
+# #############################################################################
+
+
+class Test_dassert_file_extension1 (hunitest.TestCase):
+    def test1(self) -> None:
+        """
+        Check if file has given extension.
+        """
+        extension = r".csv.gz"
+        filename = r"extensionFile.csv.gz"
+        hdbg.dassert_file_extension(filename, extension)
+
+    def test2(self) -> None:
+        """
+        Check if file has one of the given extensions.
+        """
+        extension = [r".csv.gz", r".csv"]
+        filename = r"extensionFile.csv"
+        hdbg.dassert_file_extension(filename, extension)
+    
+    def test3(self) -> None:
+        """
+        Case of filename that have incorrect extension with given extensions.
+        """
+        extension = [r".csv.gz", r".csv"]
+        filename = r"txtFile.txt"
+        with self.assertRaises(Exception) as cm:
+            hdbg.dassert_file_extension(filename, extension)
+        act = str(cm.exception)
+        exp = """
+################################################################################
+* Failed assertion *
+'.txt' in '['.csv.gz', '.csv']'
+Invalid extension '.txt' for file 'txtFile.txt'
+################################################################################
+        """
+        self.assert_equal(act, exp)


dreamcatcher0101:
Please review my code.

helpers/test/test_dbg.py
```diff
@@ -624,3 +624,42 @@ def test3(self) -> None:
...
+
+# #############################################################################
+
+
+class Test_dassert_file_extension1 (hunitest.TestCase):
+    def test1(self) -> None:
+        """
+        Check if file has given extension.
+        """
+        extension = r".csv.gz"
```
PomazkinG, line 637: why do we need `r` everywhere? can we remove?

helpers/hdbg.py
```diff
@@ -806,7 +806,7 @@ def dassert_file_extension(
     # Make sure extension starts with .
     extensions = [f".{e}" if not e.startswith(".") else e for e in extensions]
     # Check.
-    act_ext = os.path.splitext(file_name)[-1].lower()
+    act_ext = file_name[file_name.find("."):]
```
PomazkinG, line 809: an example from the real-life: typically we use also use `.` to separate pieces of information in the file name. 

e.g., `order.20231701_120057.20231701_13000.csv.gz`. here we separate timestamps with `.`. According to your code the resulting extension will be `20231701_120057.20231701_13000.csv.gz`

please modify the algo + add a separate test for `order.20231701_120057.20231701_13000.csv.gz`

dreamcatcher0101:
Hi. @PomazkinG 
Please review again.

helpers/hdbg.py
```diff
@@ -308,9 +308,13 @@ def dassert_in(
     *args: Any,
     only_warning: bool = False,
 ) -> None:
-    cond = value in valid_values
+    cond = False
+    fileExtension = act_ext = os.path.splitext(str(value))[-1].lower()
```
PomazkinG, line 312: not sure it's a good idea to modify `dassert_in()` just for checking file extensions because it's a general function (grep for it)

1. It will lead to a global regression break
2. the function is not designed for file extensions exclusively 

what I would do:
1. Modify only `dassert_file_extension()`
2. Maybe keep the existing logic
3. Add a special case: if a file name ends with `.gz` the extension has the pattern `.{extension}.gz` otherwise the pattern is `.{extension}`
   - Probably `.gz` is the only popular extension that has 2 `.`, e.g., `.csv.gz` so treating it as a special case is fine IMO
   - In simple words the idea is: extension is the set of chars after a dot, but if a file name ends with `.gz` the extension is the set of chars starting from the dot that goes before `.gz`

PomazkinG: E.g., `.csv` -> extension is `.csv`, `.csv.gz` -> extension is `.csv.gz`

DanilYachmenev:
@dreamcatcher0101 pls change PR labels accordingly when you want your PR to be reviewed
https://github.com/sorrentum/sorrentum/blob/master/docs/First_review_process.md#ask-for-reviews

helpers/hdbg.py
```diff
@@ -807,6 +807,9 @@ def dassert_file_extension(
     extensions = [f".{e}" if not e.startswith(".") else e for e in extensions]
     # Check.
     act_ext = os.path.splitext(file_name)[-1].lower()
+    if act_ext == ".gz":
+        act_ext = os.path.splitext(file_name[:-3])[-1].lower() + ".gz"
```
DanilYachmenev, line 811: Let's try to make logic a bit clearer with smth like
```
name, act_ext = os.path.splitext(file_name.lower())
if act_ext == ".gz"
    # Get a file extension that is preceding gzip extension. 
    ext = os.path.splitext(name)[-1]
    act_ext = ext + act_ext
```

helpers/test/test_dbg.py
```diff
@@ -624,3 +624,51 @@ def test3(self) -> None:
...
+        """
+        Check if file has one of the given extensions.
+        """
+        extension = [".csv.gz", ".csv"]
+        filename = "extensionFile.csv"
+        hdbg.dassert_file_extension(filename, extension)
+    
+    def test3(self) -> None:
+        """
+        Case of filename that have incorrect extension with given extensions.
```
DanilYachmenev, line 651: -> `Check that assertion is raised if filename has none of the given extensions.`

helpers/test/test_dbg.py
```diff
@@ -624,3 +624,51 @@ def test3(self) -> None:
...
+'.txt' in '['.csv.gz', '.csv']'
+Invalid extension '.txt' for file 'txtFile.txt'
+################################################################################
+        """
+        self.assert_equal(act, exp)
+
+    def test4(self) -> None:
+        """
+        Edge Case of multi dots.
+        e.g. filename = "order.20231701_120057.20231701_13000.csv.gz"
```
DanilYachmenev, line 670: -> `Check that a file with multiple dots in its name passes the assertion.`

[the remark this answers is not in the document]
dreamcatcher0101: Hi, @DanilYachmenev .
@PomazkinG asked me to add a separate test for "order.20231701_120057.20231701_13000.csv.gz", so I added it.
Is it ok to remove?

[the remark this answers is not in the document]
DanilYachmenev: Oh, I menat remove the 2nd docstring line `e.g. filename = "order.20231701_120057.20231701_13000.csv.gz"`
keep the test

helpers/hdbg.py
```diff
@@ -807,6 +807,11 @@ def dassert_file_extension(
     extensions = [f".{e}" if not e.startswith(".") else e for e in extensions]
     # Check.
     act_ext = os.path.splitext(file_name)[-1].lower()
+    if act_ext == ".gz":
+        # Get a file extension that is preceding gzip extension. 
+        # get the set of chars starting from the dot that goes before '.gz'
+        ext = os.path.splitext(file_name[:-3])[-1].lower()
+        act_ext = ext + act_ext
```
DanilYachmenev, line 814: see my suggestion for these lines
IMO `os.path.splitext(file_name[:-3])[-1].lower()` is quite vague, takes more time to comprehend

also you use `.lower()` twice here which is not DRY

helpers/hdbg.py
DanilYachmenev: https://github.com/sorrentum/sorrentum/pull/477#discussion_r1279106232

DanilYachmenev:
Press `Resolve conversation` after you implement the comment

helpers/hdbg.py
DanilYachmenev: Also pls run the tests after you implement all the changes and confirm that they pass

helpers/test/test_dbg.py
```diff
@@ -624,3 +624,51 @@ def test3(self) -> None:
...
+        filename = "txtFile.txt"
+        with self.assertRaises(Exception) as cm:
+            hdbg.dassert_file_extension(filename, extension)
+        act = str(cm.exception)
+        exp = """
+################################################################################
+* Failed assertion *
+'.txt' in '['.csv.gz', '.csv']'
+Invalid extension '.txt' for file 'txtFile.txt'
+################################################################################
```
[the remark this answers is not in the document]
dreamcatcher0101: @DanilYachmenev 
When I try to indent it on the same level where quotes start, the tab are reflected to the output and the result fails.
Is it ok to implement it as one line? like this...
exp = """\n################################################################################\n* Failed assertion *\n'.txt' in '['.csv.gz', '.csv']'\nInvalid extension '.txt' for file 'txtFile.txt'\n################################################################################\n"""
I think it's a little bit weird, but it's my solution.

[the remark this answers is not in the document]
DanilYachmenev: add `r` before right the quotes, like
```
exp = r"""
...
this should solve the problem
```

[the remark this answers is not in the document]
dreamcatcher0101: @DanilYachmenev 
I couldn't solve the problem with adding "r".
The tabs are still reflected to the output.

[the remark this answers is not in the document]
DanilYachmenev: you also need to fix the last line ->
```
self.assert_equal(act, exp, purify_text=True, fuzzy_match=True)
``` 

and use `r"""` for the output ofc

helpers/test/test_dbg.py
```diff
@@ -624,3 +625,51 @@ def test3(self) -> None:
         message 'hello world'
         """
         self.assert_equal(act, exp, purify_text=True, fuzzy_match=True)
+
+
+# #############################################################################
+
+
+class Test_dassert_file_extension1 (hunitest.TestCase):
```
PomazkinG, line 633: nit: remove space before `(...)`

helpers/test/test_dbg.py
```diff
@@ -624,3 +625,51 @@ def test3(self) -> None:
...
+            hdbg.dassert_file_extension(filename, extensions)
+        act = str(cm.exception)
+        exp = r"""
+        ################################################################################        
+        * Failed assertion *
+        '.txt' in '['.csv.gz', '.csv']'
+        Invalid extension '.txt' for file 'txtFile.txt'
+        ################################################################################
+        """
+        self.assert_equal(act, exp, purify_text=True, fuzzy_match=True)
```
PomazkinG, line 667: nit: maybe not use the `purify` switch here?

in this case we can:
1) remove the `set_user_name()`
2) the `hsystem` import

it does not seem to do its job here

devops/docker_build/create_users.sh
```diff
@@ -12,8 +12,8 @@ useradd -u 501 $OPTS user_501
 usermod -aG docker user_501
 
 # Linux users.
-# We start from id=1001, the upper bound can be changed if needed.
-for current_linux_id in {1001..1050}
+# We start from id=1000 because IDs below 1000 are usually reserved for system users and groups, the upper bound can be changed if needed.
```
PomazkinG, line 15: no need to it in this PR, the fix belongs to https://github.com/sorrentum/sorrentum/pull/465, please revert

helpers/hdbg.py
```diff
@@ -806,7 +806,11 @@ def dassert_file_extension(
     # Make sure extension starts with .
     extensions = [f".{e}" if not e.startswith(".") else e for e in extensions]
     # Check.
-    act_ext = os.path.splitext(file_name)[-1].lower()
+    name, act_ext = os.path.splitext(file_name)
```
PomazkinG, line 809: nit: remove the TODO at line 793, since your PR addresses it

helpers/hdbg.py
```diff
@@ -806,7 +806,11 @@ def dassert_file_extension(
...
+    if act_ext == ".gz":
+        # Get a file extension that is preceding gzip extension. 
+        ext = os.path.splitext(name)[-1]
+        act_ext = (ext + act_ext).lower()
```
PomazkinG, line 813: nit:
```
# Concatenate with the preceding extension, e.g., `.csv.gz`.
```

PomazkinG (commented):
Nice, only nits and we merge

DanilYachmenev (commented):
lgtm

just pls confirm that the tests pass (especially `test3` since we had problems with it)

dreamcatcher0101:
> lgtm
> 
> just pls confirm that the tests pass (especially `test3` since we had problems with it)

All the tests are passed.

PomazkinG (approved):
LGTM

Issue #479, opened 2023-07-29
Fix contributor doc file path in ReadMe 


sonaalKant:
In README change the path ```docs/Contributor_scorecard.md``` to ```docs/Contributor_Scorecard.md```

PR #480, opened 2023-07-29
fix file path name in readme

sonaalKant:
Issue #479 

@DanilYachmenev The Contributor_scorecard path was broken which I fixed now.


DanilYachmenev:
Nice call, thx, merging

PR #495, opened 2023-08-01
improve multiindex_df_info() and add unittest

critrange:


helpers/hpandas.py
```diff
@@ -1838,6 +1838,14 @@ def multiindex_df_info(
         "columns_level1=%s" % list_to_str(columns_level1, **list_to_str_kwargs)
     )
     ret.append("rows=%s" % list_to_str(rows, **list_to_str_kwargs))
+    # Check if the index is a DatetimeIndex and get more info
```
PomazkinG, line 1841: remove

helpers/hpandas.py
```diff
@@ -1838,6 +1838,14 @@ def multiindex_df_info(
...
+    if isinstance(df.index, pd.DatetimeIndex):
+        start_timestamp = df.index.min()
```
PomazkinG, line 1843:
```
# Display timestamp info.
```

helpers/hpandas.py
```diff
@@ -1838,6 +1838,14 @@ def multiindex_df_info(
...
+        end_timestamp = df.index.max()
+        frequency = pd.infer_freq(df.index)
```
PomazkinG, line 1845: here we are trying to infer frequency 

1. can we check if there is an assigned frequency?
2. how do we handle a situation when freq cannot be inferred?

helpers/test/test_hpandas.py
```diff
@@ -3096,3 +3096,112 @@ def test1(self)-> None:
...
+        df = pd.DataFrame(data, index=non_datetime_index, columns=index)
+        return df
+
+    def test1(self) -> None:
+        """
+        Test DataFrame with a datetime index.
+        """
+        df = self.get_multiindex_df_with_datetime_index()
+        act = hpandas.multiindex_df_info(df)
+        self.assertIn("shape=2 x 4 x 5", act)
```
PomazkinG, line 3184: can we just freeze the entire `act`? I mean compare the entire `act` string with an expected value using `self.assert_equal()`

PomazkinG (commented):
Nice!

helpers/hpandas.py
PomazkinG: maybe `None`

helpers/hpandas.py
```diff
@@ -1838,6 +1838,15 @@ def multiindex_df_info(
         "columns_level1=%s" % list_to_str(columns_level1, **list_to_str_kwargs)
     )
     ret.append("rows=%s" % list_to_str(rows, **list_to_str_kwargs))
+    if isinstance(df.index, pd.DatetimeIndex):
+        # Display timestamp info.
+        start_timestamp = df.index.min()
+        end_timestamp = df.index.max()
+        frequency = pd.infer_freq(df.index)
```
PomazkinG, line 1845: I would do:
```
frequency = df.index.freq
if frequency is None:
   # Try to infer frequency.
   frequency = pd.infer_freq(df.index)
```

in other words the flow is:
1) try to get an assigned frequency using `index.freq`
2) if it's not defined try to infer
3) if pandas cannot infer it will return None -> no need to handle it since pandas natively handles it

helpers/test/test_hpandas.py
```diff
@@ -3096,3 +3096,143 @@ def test1(self)-> None:
...
+               "shape=2 x 4 x 5",
+               "columns_level0=2 ['asset1', 'asset2']",
+               "columns_level1=4 ['close', 'high', 'low', 'open']",
+               "rows=5 ['2022-01-01 21:01:00+00:00', '2022-01-01 21:02:00+00:00', '2022-01-01 21:04:00+00:00', '2022-01-01 21:04:30+00:00', '2022-01-01 21:06:00+00:00']",
+               "start_timestamp=2022-01-01 21:01:00+00:00",
+               "end_timestamp=2022-01-01 21:06:00+00:00",
+               "frequency=None"
+            ]
+        )
+        self.assert_equal(act, exp)
```
PomazkinG, line 3222: can we just use the multiline string like so:
```
exp = """
some text
more text
and even more text
"""
```

PomazkinG: add also fuzzy_match=True to `assert_equal()`

Issue #493, opened 2023-08-01
LIN: Rename the gsheets in doc and code


gpsaggese:
We want to rename the gsheets
<img width="741" alt="image" src="https://github.com/sorrentum/sorrentum/assets/33238329/20c8d8ba-3700-4cb5-88ff-1a824b136c14">

so that the flow is clear in both the notebooks and in the documentation (https://docs.google.com/document/d/1dfAVMJZtHtQ3sX7UHwpXNAWPiJs17w0V8FlsOI7a32M/edit)
like:

- Sn_search5.step1.export_search
- Sn_search5.step2.export_search.filtered
- Sn_search5.step3.export_profile
- Sn_search5.step4.export_profile.filtered

yiyunlei:
Rename them as shown below to make them more concise and clear:
<img width="862" alt="image" src="https://github.com/sorrentum/sorrentum/assets/42398651/e666861e-e3b0-4cd2-9ab5-032a16e1edec">

Issue #496, opened 2023-08-01
Refine Google Sheet Flow


yiyunlei:
- Goal: Reduce manual work.

- Plan: 
    - Create a folder to store all results -> generate the Google Sheet file in this folder. If the file already exists, simply use it. (Each folder has 4 files.)
    - Structure:

      ```
      linkein_data/
          sn_search5/
               sn_search5.step1.search_export
               sn_search5.step2.search_export_filtered
               sn_search5.step3.profile_export
               sn_search5.step4.profile_export_filtered
          sn_search6/
               sn_search6.step1.search_export
               sn_search6.step2.search_export_filtered
               sn_search6.step3.profile_export
               sn_search6.step4.profile_export_filtered
      ```
- Steps:
    - Create the sn_searchN folder by using Google Drive folder API.(TODO)
      Resource: https://developers.google.com/drive/api/guides/folder
    - Create four empty gsheet files in this folder. (Implemented)

yiyunlei:
- ETA: Next Monday(2023-08-07)
  - Refine code and code style.
  - Add function to notebooks.

yiyunlei:
- ETA: Wednesday(2023-08-09)
- Original ETA: Monday (2023-08-07)
- Reason: Need more time for the previous issue and this issue.
  - Refine code and code style.
  - Add the create folder function to notebooks and clean notebooks.

yiyunlei:
- ETA: Today(2023-08-09)
    - Clean code and submit PR.

PR #505, opened 2023-08-02
Sorr Task493 Rename gsheets name

yiyunlei:
for #493

### DIFF
--- a/linkedin/notebooks/process_profile_export.ipynb
+++ b/linkedin/notebooks/process_profile_export.ipynb
@@ -10,24 +10,15 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 10,
    "id": "e77e9172",
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2023-07-26T18:40:32.272328Z",
-     "start_time": "2023-07-26T18:40:29.730965Z"
+     "end_time": "2023-08-02T21:12:56.360085Z",
+     "start_time": "2023-08-02T21:12:56.340815Z"
     }
    },
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)\n",
-      "<jemalloc>: (This is the expected behaviour if you are running under QEMU)\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
     "import gspread_pandas\n",
     "import pandas as pd"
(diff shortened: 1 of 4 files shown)


linkedin/notebooks/process_profile_export.ipynb
```diff
@@ -373,7 +373,7 @@
     }
    ],
    "source": [
-    "spreadsheet_name = \"sn_search5.profile_export.gsheet\"\n",
+    "spreadsheet_name = \"3_profile_export_sns5\"\n",
```
gpsaggese, line 376: All the info it's there. The question is how we want to organize the data

IMO the best organization is a single dir with all the data inside (unless we want to separate by search). In this case the dir layout is:

```
linked_in_data/
     sn5.step1.profile_export
     sn5.step2.profile_export_filtered
```
since then the alphabetical order matches how we use the data:
- all the data for a search together
- then the phases in order (with a human readable explanation of the phase later)

Can you use this organization and also add this explanation to the How to gdoc?

linkedin/notebooks/process_search_export.ipynb
```diff
@@ -438,7 +438,7 @@
     }
    ],
    "source": [
-    "spreadsheet_name = \"sn_search5.search_export.gsheet\"\n",
+    "spreadsheet_name = \"1_search_export_sns5\"\n",
```
gpsaggese, line 441: sns5 should be factored out in a string like:

```
search_name = "sns5"
...
spreadsheet_name = f"{search_name}.step1.search_export"
```

gpsaggese, line 441: let's use the same name for the same things, so spreadsheet_name -> gsheet_name

linkedin/notebooks/process_profile_export.ipynb
yiyunlei: OK,no problem.

yiyunlei: I think it's better to separate by search like below(put different search results in different folders.), it would be more clear:
```
linkin_data/
    sn_search5/
         sn5.step1.search_export
         sn5.step2.search_export_filtered
         sn5.step3.profile_export
         sn5.step4.profile_export_filtered
    sn_search6/
         sn6.step1.search_export
         sn6.step2.search_export_filtered
         sn6.step4.profile_export
         sn6.step5.profile_export_filtered

```

gpsaggese: 1) It's ok with me, but I understood that creating dirs under Google drive was not simple. if that's not a problem, sure more structure is better

2) In any case let's do the same name for dirs and prefix (so the dir structure is just to avoid visual noise)
```
sn_search5/
     sn_search5.step1.search_export
```

yiyunlei: I will rename gsheets first, and see if I can create the folder. (It would be beneficial for the entire flow.)

yiyunlei: I found the corresponding API for it here: https://developers.google.com/drive/api/guides/folder, so it's practical. I've also updated the description for issue #496.

yiyunlei:
I've also updated the Google doc [LinkedIn - Automation Process
](https://docs.google.com/document/d/1dfAVMJZtHtQ3sX7UHwpXNAWPiJs17w0V8FlsOI7a32M/edit#heading=h.lkq5uolbnxir)

gpsaggese (commented):
Good work

Issue #502, opened 2023-08-02
Disable timeout for tests when running outside of CI


gpsaggese:
Thanks. Took a glance. Maybe for the timeouts [Category 1] we can mark them as `slow` instead.

_Originally posted by @LibertasSpZ in https://github.com/sorrentum/sorrentum/issues/283#issuecomment-1656294100_

The proposed solution for this is in
           
https://docs.google.com/document/d/1Qm54LwNRlBwzYroDGviW32NevK_V8L4Sufsg5u6IKwI/edit#heading=h.6bhp1lld1a4v

We want to make the timeout in `pytest` larger when we are running the tests outside of GH actions / CI, when we build the command line option for `run_fast_tests`

In practice, it's something like in `_build_run_command_line`
```
if is_ck_server() or CI:
        # Make the timeout much larger.
        pytest_opts_tmp.append("-o timeout_func_only=true")
        timeout_secs *= 10
        pytest_opts_tmp.append(f"--timeout {timeout_in_sec}")
```

LibertasSpZ:
So after merging #500 today we have the following in `helpers/lib_tasks_pytest.py`

   ```
   skip_ck_infra_tests = not hserver.is_dev_ck()
      if skip_ck_infra_tests:
          _LOG.warning("Skipping CK infra related tasks")
          skipped_tests += " and not requires_ck_infra"
          timeout_in_sec *= 10
   ```


I believe that already disables [or significantly extends] timeouts for fast tests running outside of ck infrastructure.

gpsaggese:
Done

Issue #501, opened 2023-08-02
Fix each cause of issues when running regressions


gpsaggese:
From #283

Create a test list for each class of broken tests based on
https://docs.google.com/document/d/1Qm54LwNRlBwzYroDGviW32NevK_V8L4Sufsg5u6IKwI/edit#heading=h.6bhp1lld1a4v

We want to create a GH issue for each of the issues we need to propose a solution (e.g., fix so that it works outside our server, disabled since it can't run outside CK infra)

At the end the regressions should pass with a subset of the tests skipped only when running CK infra

LibertasSpZ:
For some reason I get access denied for that doc now.

gpsaggese:
Let's add a switch to force a test list to run. This should be a separate issue.

The workflow is going to be like:
- enable a subset of tests from command line (e.g., `requires_ck_infra` that would be skipped otherwise)
- classify failures like in the gdoc https://docs.google.com/document/d/1Qm54LwNRlBwzYroDGviW32NevK_V8L4Sufsg5u6IKwI/edit
- file an issue for each root cause
- decide why it fails, if it can be fixed, and how
   - e.g., the timeout issue is fixed by increasing the timeout)
   - e.g., `Got permission denied while trying to connect to the Docker daemon socket at` we will mark it as `requires_dind` and skip them (so is_laptop implies `requires_dind`)

Does it make sense?

LibertasSpZ:
Yes. I'll start by creating a branch and adding a switch to allow only the `requires_ck_infra` to run, and create a new issue for each type of failure / error.

Issue #500, opened 2023-08-02
Disable failing tests with `requires_ck_infra`


gpsaggese:
From #282 and #283

We want to disable all the failing tests so we can run the regressions in Sorrentum

The goal is to run on your laptop `> i run_fast_tests` and complete without errors / failures skipping all the broken tests 
This should ensure that GitHub actions in https://github.com/sorrentum/sorrentum/blob/master/.github/workflows/fast_tests.yml (which runs `run: invoke run_fast_tests`) completes without problem

1) Add the marker `requires_ck_infra` to the broken tests
2) Add a switch to `invoke run_fast_tests` (`--not-ck-infra`)
3) Automatically let the system decide if we are inside the `ck_infra` or not inside the invoke target `run_fast_tests`
    - There is draft PR in SorrTask283_Create_test_list_to_run_with_Sorrentum_2
    - You can use `helpers.hserver.is_dev_ck()` to detect if we are running on a CK server or not

FYI @jsmerix

PR #521, opened 2023-08-05
SorrTask501_Fix_each_cause_of_issues_when_running_regressions

LibertasSpZ:
Run fast tests on only marked as `require_ck_infra`. Progress in #501 .                            

[1] `timeout > 5.0s` is fully resolved by the `lib_tasks_pytest.py` in #500, now merged to master. See full report attached in [2]; see also comments in #502. #502 is thus recommended to be closed together with #501 .

[2] Only 2 types of errors remain full report attached here [ck_infra_out_Aug042023.txt](https://github.com/sorrentum/sorrentum/files/12265398/ck_infra_out_Aug042023.txt). I don't see a "docker in docker", but please feel free to double check. In the full report, "Test Session Starts": line 181; "FAILURES": Line 357. 

The 2 types:

   Type 1: requires aws.
       Type 1 proposed solution: file a new issue replacing #501 . In a new branch, add switch ` and not requires_aws` in `lib_tasks_pytest.py`.
       Note: I already marked tests that `requires_aws` in #500 which is now merged in master, so we just need to test the switch works.

   Type 2: FUZZY ACTUAL v. FUZZY EXPECTED
       Type 2 proposed solution: file a new issue replacing #501 . In a new branch, reduce the number of digits of output e.g. in `tau`, as proposed in [the google doc](https://docs.google.com/document/d/1Qm54LwNRlBwzYroDGviW32NevK_V8L4Sufsg5u6IKwI/edit#heading=h.6bhp1lld1a4v).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -40,6 +40,19 @@ if not hasattr(hut, "_CONFTEST_ALREADY_PARSED"):
     # Add custom options.
     def pytest_addoption(parser: Any) -> None:
         parser.addoption(
+            "--run_only_ck_infra_tests",
+            action="store_true",
+            default=False,
+            help="Running only CK infra related tasks",
+        )
+        #keep both switch to display different warning msgs.
+        parser.addoption(
+            "--skip_ck_infra_tests",
+            action="store_true",
+            default=False,
+            help="Skipping CK infra related tasks",
+        )
+        parser.addoption(
             "--update_outcomes",
             action="store_true",
             default=False,
@@ -73,6 +86,9 @@ if not hasattr(hut, "_CONFTEST_ALREADY_PARSED"):
             help="Stage of the image to test against",
         )
 
+    def _update_marker_expr(marker_expr: str, added_marker: str) -> str:
+        return f"{marker_expr} and {added_marker}" if marker_expr else added_marker
+
     def pytest_collection_modifyitems(config: Any, items: Any) -> None:
         _ = items
         import helpers.henv as henv
@@ -82,6 +98,14 @@ if not hasattr(hut, "_CONFTEST_ALREADY_PARSED"):
             print(henv.get_system_signature()[0])
         except:
             print(f"\n{_WARNING}: Can't print system_signature")
+        if config.getoption("--run_only_ck_infra_tests"):
+            print(f"\n{_WARNING}: Running only CK infra related tasks")
+            marker_expr = config.option.markexpr
+            config.option.markexpr = _update_marker_expr(marker_expr, "requires_ck_infra")
+        if config.getoption("--skip_ck_infra_tests"):
+            print(f"\n{_WARNING}: Skipping CK infra related tasks")
+            marker_expr = config.option.markexpr
+            config.option.markexpr = _update_marker_expr(marker_expr, "not requires_ck_infra")
         if config.getoption("--update_outcomes"):
             print(f"\n{_WARNING}: Updating test outcomes")
             hut.set_update_tests(True)
--- a/conftest.py
+++ b/conftest.py
@@ -40,19 +40,6 @@ if not hasattr(hut, "_CONFTEST_ALREADY_PARSED"):
     # Add custom options.
     def pytest_addoption(parser: Any) -> None:
         parser.addoption(
-            "--run_only_ck_infra_tests",
-            action="store_true",
-            default=False,
-            help="Running only CK infra related tasks",
-        )
-        #keep both switch to display different warning msgs.
-        parser.addoption(
-            "--skip_ck_infra_tests",
-            action="store_true",
-            default=False,
-            help="Skipping CK infra related tasks",
-        )
-        parser.addoption(
             "--update_outcomes",
             action="store_true",
             default=False,
@@ -86,8 +73,6 @@ if not hasattr(hut, "_CONFTEST_ALREADY_PARSED"):
             help="Stage of the image to test against",
         )
 
-    def _update_marker_expr(marker_expr: str, added_marker: str) -> str:
-        return f"{marker_expr} and {added_marker}" if marker_expr else added_marker
 
     def pytest_collection_modifyitems(config: Any, items: Any) -> None:
         _ = items
@@ -98,14 +83,6 @@ if not hasattr(hut, "_CONFTEST_ALREADY_PARSED"):
             print(henv.get_system_signature()[0])
         except:
             print(f"\n{_WARNING}: Can't print system_signature")
-        if config.getoption("--run_only_ck_infra_tests"):
-            print(f"\n{_WARNING}: Running only CK infra related tasks")
-            marker_expr = config.option.markexpr
-            config.option.markexpr = _update_marker_expr(marker_expr, "requires_ck_infra")
-        if config.getoption("--skip_ck_infra_tests"):
-            print(f"\n{_WARNING}: Skipping CK infra related tasks")
-            marker_expr = config.option.markexpr
-            config.option.markexpr = _update_marker_expr(marker_expr, "not requires_ck_infra")
         if config.getoption("--update_outcomes"):
             print(f"\n{_WARNING}: Updating test outcomes")
             hut.set_update_tests(True)
--- a/conftest.py
+++ b/conftest.py
@@ -73,7 +73,6 @@ if not hasattr(hut, "_CONFTEST_ALREADY_PARSED"):
             help="Stage of the image to test against",
         )
 
-
     def pytest_collection_modifyitems(config: Any, items: Any) -> None:
         _ = items
         import helpers.henv as henv
--- a/helpers/lib_tasks_pytest.py
+++ b/helpers/lib_tasks_pytest.py
@@ -145,8 +145,8 @@ def _build_run_command_line(
     # Detect if we are running on a laptop.
     skip_ck_infra_tests = not hserver.is_dev_ck()
     if skip_ck_infra_tests:
-        _LOG.warning("Skipping CK infra related tasks")
-        skipped_tests += " and not requires_ck_infra"
+        _LOG.warning("running only CK_infra-related tasks to find cause of err")
+        skipped_tests += " and requires_ck_infra"
         timeout_in_sec *= 10
     if custom_marker != "":
         pytest_opts_tmp.append(f'-m "{custom_marker} and {skipped_tests}"')
--- a/helpers/lib_tasks_pytest.py
+++ b/helpers/lib_tasks_pytest.py
@@ -143,8 +143,8 @@ def _build_run_command_line(
     skipped_tests = _select_tests_to_skip(test_list_name)
     timeout_in_sec = _TEST_TIMEOUTS_IN_SECS[test_list_name]
     # Detect if we are running on a laptop.
-    outside_ck = not hserver.is_dev_ck()
-    if outside_ck:
+    is_outside_ck_infra = not hserver.is_dev_ck()
+    if is_outside_ck_infra:
         # Since we are running outside the CK server we increase the duration
         # of the timeout, since the thresholds are set for the CK server.
         timeout_in_sec *= 10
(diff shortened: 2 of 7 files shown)


helpers/lib_tasks_pytest.py
```diff
@@ -145,8 +145,8 @@ def _build_run_command_line(
     # Detect if we are running on a laptop.
     skip_ck_infra_tests = not hserver.is_dev_ck()
     if skip_ck_infra_tests:
-        _LOG.warning("Skipping CK infra related tasks")
-        skipped_tests += " and not requires_ck_infra"
+        _LOG.warning("running only CK_infra-related tasks to find cause of err")
```
gpsaggese, line 148: - The switch `skip_ck_infra_tests` now behaves the opposite of what the name says
- We want another switch that is called `run_only_ck_infra_tests` that is passed from command line

helpers/lib_tasks_pytest.py
```diff
@@ -145,8 +145,8 @@ def _build_run_command_line(
...
+        skipped_tests += " and requires_ck_infra"
         timeout_in_sec *= 10
```
gpsaggese, line 150: We need a comment here.
```
# Since we are running outside the CK server we increase the duration of the timeout, since the thresholds are set for the CK server.
```

helpers/lib_tasks_pytest.py
gpsaggese, line 148: IMO the previous message is better. We should explain what's happening and not what it is used for (mechanism vs policy)

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -171,12 +181,21 @@ def test_run_fast_tests5(self) -> None:
             tee_to_file,
             n_threads,
         )
-        exp = (
-            'pytest -m "not slow and not superslow" . '
-            "-o timeout_func_only=true --timeout 5 --reruns 2 "
-            '--only-rerun "Failed: Timeout" -n 1 2>&1'
-            " | tee tmp.pytest.fast_tests.log"
-        )
+        skip_ck_infra_tests = not hserver.is_dev_ck()
```
gpsaggese, line 184: Let's factor this out. We simply abhor copy-paste. We obey to DRY principle like a law of nature.

gpsaggese:
Good explanation of everything. Let's have a quick PP on Mon so I can explain you the specs a bit better

LibertasSpZ:
Added the `run_only_ck_infra_tests` switch as required. 

I have trouble letting `invoke` understand the switches though. When I run `i run_fast_tests --run_only_ck_infra_tests` it fails to recognize the switch; `i  run_fast_tests(pytest_opts = "--run_only_ck_infra_tests")` gives me syntax error. 

How do I pass the switch to `run_fast_tests` as the argument `pytest_opts` using `invoke` from command line?

conftest.py
```diff
@@ -39,6 +39,19 @@ def populate_globals(capsys):
...
+            default=False,
+            help="Running only CK infra related tasks",
+        )
+        #keep both switch to display different warning msgs.
+        parser.addoption(
+            "--skip_ck_infra_tests",
+            action="store_true",
+            default=False,
+            help="Skipping CK infra related tasks",
+        )
```
gpsaggese, line 54: Good job digging in the codebase and finding a solution.
Your approach is for adding command line opts to pytest.

To add it to `invoke` you need to add a param to the task, which automatically adds command line opts
```
@task
def run_fast_tests(  # type: ignore
    ctx,
    stage="dev",
    version="",
    pytest_opts="",
    skip_submodules=False,
    coverage=False,
    collect_only=False,
    tee_to_file=False,
    n_threads="serial",
    git_clean_=False,
):
```
For some context you can look at
- https://www.pyinvoke.org/
- We have some doc here docs/Development_workflow.md

conftest.py
```diff
@@ -73,6 +86,9 @@ def pytest_addoption(parser: Any) -> None:
             help="Stage of the image to test against",
         )
 
+    def _update_marker_expr(marker_expr: str, added_marker: str) -> str:
```
gpsaggese, line 89: In practice you don't need to change `conftest.py`

helpers/lib_tasks_pytest.py
```diff
@@ -143,10 +143,10 @@ def _build_run_command_line(
     skipped_tests = _select_tests_to_skip(test_list_name)
     timeout_in_sec = _TEST_TIMEOUTS_IN_SECS[test_list_name]
     # Detect if we are running on a laptop.
-    skip_ck_infra_tests = not hserver.is_dev_ck()
-    if skip_ck_infra_tests:
-        _LOG.warning("Skipping CK infra related tasks")
-        skipped_tests += " and not requires_ck_infra"
+    outside_ck = not hserver.is_dev_ck()
```
gpsaggese, line 146: -> is_outside_ck_infra

conftest.py
LibertasSpZ: Thanks. That was helpful!

LibertasSpZ:
Thanks for the explanations in the closed conversations. Now

   ```
   i run_fast_tests -r Yes
   ```
runs the desired tests marked with `requires_ck_infra and not slow and not superslow`, generating 58 errors in 2 types as discussed above.

conftest.py
```diff
@@ -73,6 +73,7 @@ def pytest_addoption(parser: Any) -> None:
             help="Stage of the image to test against",
         )
 
+
```
gpsaggese, line 76: Pls revert this one

(Always take another look at the PR with fresh eyes to catch these small issues)

helpers/lib_tasks_pytest.py
```diff
@@ -143,10 +143,10 @@ def _build_run_command_line(
     skipped_tests = _select_tests_to_skip(test_list_name)
     timeout_in_sec = _TEST_TIMEOUTS_IN_SECS[test_list_name]
     # Detect if we are running on a laptop.
```
gpsaggese, line 145: Note that the comment talks about laptop but the code about a server. Let's be clear.

```
# Detect if we are running on a CK dev server or a laptop outside the CK infra.
```

helpers/lib_tasks_pytest.py
```diff
@@ -364,6 +364,7 @@ def run_tests(  # type: ignore
 def run_fast_tests(  # type: ignore
     ctx,
     stage="dev",
+    require_ck="",
```
gpsaggese, line 367: let's put it as last param in this case.

There are "groups" of params that we want to keep together (you didn't know this)

gpsaggese, line 367: let's make it boolean `=False`. invoke has a way to handle boolean params through the config.
(e.g., look at the code around, and you can see that we use boolean params -> then you can infer that there must be something special -> check the doc)

helpers/lib_tasks_pytest.py
```diff
@@ -374,9 +375,10 @@ def run_fast_tests(  # type: ignore
     git_clean_=False,
 ):
     """
-    Run fast tests.
+    Run fast tests. check `gh auth status` before invoking to avoid auth errors.
 
     :param stage: select a specific stage for the Docker image
+    :param require_ck: select whether to run tests that requires_ck_infra.
```
gpsaggese, line 381: -> `requires_ck_infra`

Better to have the most expressive name since this is exposed to the user.
At this point let's match the internal name so there is one less symbol to remember

gpsaggese, line 381: Of course keep the params in order

helpers/lib_tasks_pytest.py
```diff
@@ -389,7 +391,14 @@ def run_fast_tests(  # type: ignore
     """
     hlitauti.report_task()
     test_list_name = "fast_tests"
-    custom_marker = ""
+    if require_ck == "Yes":
+        _LOG.warning("Running ck infra related tests ONLY")
+        custom_marker = "requires_ck_infra"
```
gpsaggese, line 396: Good move to reuse the existing mechanisms

helpers/lib_tasks_pytest.py
```diff
@@ -389,7 +391,14 @@ def run_fast_tests(  # type: ignore
...
+    elif require_ck == "No":
+        _LOG.warning("Skipping ck infra related tests")
```
gpsaggese, line 398: no need to give a warning for a normal behavior

helpers/test/test_git.py
```diff
@@ -214,6 +214,9 @@ def test_get_task_prefix_from_repo_short_name1(self) -> None:
 
 #TODO(shaopengz): This test hangs in hgit.is_in_amp_as_supermodule().
```
gpsaggese, line 215: Are you using the linter? It should replace automatically

```
#TODO
```

with 

```
# TODO
```

helpers/test/test_git.py
```diff
@@ -214,6 +214,9 @@ def test_get_task_prefix_from_repo_short_name1(self) -> None:
...
 @pytest.mark.requires_ck_infra
+#TODO(shaopengz): To test requires_ck_infra and avoid hang, labeling
```
gpsaggese, line 217: Then you should mark it as `pytest.mark.skip` since it hangs and broken, rather than slow

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -21,6 +21,17 @@
 
 
 class Test_build_run_command_line1(hunitest.TestCase):
+    
+    def _compute_timeout_depending_on_ck_infra(self) -> int:
```
gpsaggese, line 25: Correct approach.

When we start using mocking all this will become simpler.

You can start take a look at `https://docs.python.org/3/library/unittest.mock.html`
docs/Unit_tests.md:53:- [Mocking](#mocking)

gpsaggese (commented):
Very close.

All the solutions you've found are valid, but often there is a "simpler" solution to those problems that you don't know (yet). Just try to understand the logic of why we do things in a certain way and look at the code around.

The first few reviews are always difficult.
Take a look at docs/Code_review.md with some suggestions

helpers/lib_tasks_pytest.py
LibertasSpZ: Thanks for the review, @gpsaggese . I really appreciate it.

I'll implement most of the other changes; however here I changing it to bool makes the behavior different. Right now the switch has 3 behaviors: 
(1) if "Yes", the markers will eventually be `requires_ck_infra and not slow and not superslow`; 
(2) if "No", the markers will eventually be `not requires_ck_infra and not slow and not superslow`; 
(3) if anything else, including empty string, the markers will be `not slow and not superslow`, which to my understanding is the real normal behavior. [This also justifies the warning "skipping ck infra related tests" in the "No" branch, (2) above]. A boolean cannot provide 3 behaviors.

I agree "Yes" and "No" is a little confusing; maybe we can change the strings to "O", "S" and empty, where "O" stands for "Only requires_ck_infra", "S" for "Skipping requires_ck_infra", empty/others for not making a choice at requires_ck_infra.

gpsaggese: Good point. I didn't understand all the nuances here.

In these cases, we usually would add two options
```
skip_requires_ck_infra_tests
only_requires_ck_infra_tests
```
You can detect the invalid condition 2 booleans encode with
```
hdbg.dassert(not (skip_requires_ck_infra_tests and only_requires_ck_infra_tests),
    msg="You can't use at the same time --skip_requires_ck_infra_tests and --only_requires_ck_infra_tests")
```

In the next PRs, we will generalize these switches to any marker (e.g., `--only_marker requires_AWS`, or skip those ones).

Does it make sense?

helpers/test/test_lib_tasks_pytest.py
LibertasSpZ: Will check out soon. Thanks for the advice!

helpers/lib_tasks_pytest.py
LibertasSpZ: Yes, this makes sense, and seems to be a clearer approach. Thanks @gpsaggese .

I'll implement the `skip_` and `only_` for `requires_ck_infra` in the next PR, and use similar approach for e.g. `requires_AWS` and other markers in future PRs.

helpers/test/test_git.py
LibertasSpZ: linted

Issue #504, opened 2023-08-02
Add people to Telegram group


samarth9008:
- We want to add and remove people to the telegram group
- Create a python script for adding and removing people from the group.
- Probably you will need to create a telegram bot.
- Add that bot as admin to that group
- The script will control the bot which add/remove the people.
- `BOT API TOKEN` should be fetched from the environment variable since we need to keep that secret.
- The script should have interface like this
   `process_telegram.py --add/remove --username <USERNMAME> --groupid <Id of the group to add to>`
- It will be inside `dev_scripts` folder.
- Make sure to create a detailed MD file so that we can follow the same steps and use it for our purpose.
- Let us know if you have any questions.

FYI @DanilYachmenev @gpsaggese

samarth9008:
Assigning myself but will re-route.

hhxjqm:
Do I need to provide a method about how to get group ID?

samarth9008:
Yes

hhxjqm:
From what I searched online, you can't get user id directly from username or mobile number

https://stackoverflow.com/questions/31078710/how-to-obtain-telegram-chat-id-for-a-specific-user

PR #523, opened 2023-08-08
SorrTask504_Add_people_to_Telegram_group

hhxjqm:
#504

### DIFF
--- a/dev_scripts/get_telegram_groupid.py
+++ b/dev_scripts/get_telegram_groupid.py
@@ -1,0 +1,37 @@
+"""
+- Open Telegram and navigate to the group for which you want to get the ID.
+
+- Add the Telegram bot that you created as an admin to the groups.
+
+- Once the bot is added to the group, send any message in the group. 
+
+to run outside a Docker container:
+```
+> python get_telegram_groupid.py
+```
+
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@
 .ipynb_checkpoints
 __pycache__
 .dockerignore
+.env
```
DanilYachmenev, line 7: why this change?

samarth9008: You can try doing this. In that way you will not have to create a new `.env` file

```
import os
value = os.environ.get("MY_VARIABLE")
```

dev_scripts/get_telegram_groupid.py
```diff
@@ -18,26 +18,27 @@
 import requests
```
samarth9008, line 18: Can't we just merge this script into the main one? Like it calls the function to get `group_id` and pass that id as arg to add and remove people from the group.

hhxjqm: @samarth9008 We have to know the groupid first, then we can call the main script.

Maybe I can do this:
```bash
process_telegram.py --action get
```
To get the group id

Issue #518, opened 2023-08-03
Add mocking to isolate the deps for ck dev servers


gpsaggese:
Ok for now but we'll do mocking to isolate the server deps.

_Originally posted by @gpsaggese in https://github.com/sorrentum/sorrentum/pull/506#discussion_r1283775643_

gpsaggese:
We can work together on this on the next PP

Issue #282, opened 2023-05-25
Enable fast tests action


jsmerix:

jsmerix:
@gpsaggese can you add me as Admin on Sorrentum ( Settings > Collaborators and teams > Manage Access > Set Role to Admin). In order to make actions work we need to set the same set of secrets. Unless we have an enterprise package I believe we cannot set secress across organizations so it has to be done on a per repo basis.

gpsaggese:
Done

- Do we have enterprise package in CK? If so we can upgrade here too to make our life easier
- There is a Python flow set up by Nikola to set all the secrets in one shot. I've used it a few times (it's documented in a gdoc)

jsmerix:
> Do we have enterprise package in CK? If so we can upgrade here too to make our life easier

Enterprise is 20$/user/month, and right out of the bat I am not sure if it still would be possible to set cross-account secrets since cmamp repo is within user `cryptomtc` and `sorrentum` is within `sorrentum`

> There is a Python flow set up by Nikola to set all the secrets in one shot. I've used it a few times (it's documented in a gdoc)

I'll try to find and use it.

gpsaggese:
> Enterprise is 20$/user/month, and right out of the bat I am not sure if it still would be possible to set cross-account secrets since cmamp repo is within user cryptomtc and sorrentum is within sorrentum

Good point. Cost is prohibitive. The easiest way is to use the script pointed by @PomazkinG to set the secrets in both repos.

jsmerix:
We have got to a checkpoint of fast tests enabled but we need to resolve test failure (most of them due to conflicts within repos) https://github.com/sorrentum/sorrentum/pull/325

jsmerix:
> @jsmerix anything else stopping from enabling the regressions?

Nothing else, this order and list of actions seems correct.

LibertasSpZ:
Since checkpoint in #524 / #527 says all fast tests passed, would this and #283 be closed?

jsmerix:
Fast tests are not enabled https://github.com/sorrentum/sorrentum/actions/runs/6229593355/job/16908348181

PR #524, opened 2023-08-08
SorrTask518_Add_mocking_to_isolate_the_deps_for_ck_dev_servers

LibertasSpZ:
Finished #518 .

To test, run `i run_mock_tests`. It runs the 10 tests modified in `helpers/test/test_lib_tasks_pytest.py` and passes all of them.

By the way the 3 failed tests on check points are 2 gh auth issues and 1 numerical issue which I guess can be @pytest.mark.skip ed.

### DIFF
--- a/helpers/lib_tasks_pytest.py
+++ b/helpers/lib_tasks_pytest.py
@@ -359,35 +359,6 @@ def run_tests(  # type: ignore
     return rc
 
 
-@task
-def run_mock_tests(ctx):
-    """
-    Only run the 10 tests in helpers/test/test_lib_tasks_pytest.py that uses mock.
-    """
-    hlitauti.report_task()
-    test_path = "helpers/test/test_lib_tasks_pytest.py::Test_build_run_command_line1"
-    cmd = "pytest"
-    for num in [1, 2, 5, 6, 7]:
-        for version in ["t", "f"]:
-            cmd += f" {test_path}::test_run_fast_tests{num}{version} "
-    base_image = ""
-    # We need to add some " to pass the string as it is to the container.
-    #######cmd = f"'{cmd}'"
-    # We use "host" for the app container to allow access to the database
-    # exposing port 5432 on localhost (of the server), when running dind we
-    # need to switch back to bridge. See CmTask988.
-    extra_env_vars = ["NETWORK_MODE=bridge"]
-    docker_cmd_ = hlitadoc._get_docker_compose_cmd(
-    base_image, "dev", "", cmd, extra_env_vars=extra_env_vars
-    )
-    _LOG.info("cmd=%s", docker_cmd_)
-    # We can't use `hsystem.system()` because of buffering of the output,
-    # losing formatting and so on, so we stick to executing through `ctx`.
-    rc: Optional[int] = hlitadoc._docker_cmd(ctx, docker_cmd_)
-    return rc
-
-
-
 # TODO(gp): Pass a test_list in fast, slow, ... instead of duplicating all the code CmTask #1571.
 @task
 def run_fast_tests(  # type: ignore
--- a/helpers/lib_tasks_pytest.py
+++ b/helpers/lib_tasks_pytest.py
@@ -359,6 +359,19 @@ def run_tests(  # type: ignore
     return rc
 
 
+@task
+def run_mock_tests(ctx):
+    """
+    Only run the 10 tests in helpers/test/test_lib_tasks_pytest.py that uses mock.
+    """
+    test_path = "helpers/test/test_lib_tasks_pytest.py::Test_build_run_command_line1"
+    for num in [1, 2, 5, 6, 7]:
+        for version in ["t", "f"]:
+            ctx.run(f"pytest {test_path}::test_run_fast_tests{num}{version}")
+    return 0
+
+
+
 # TODO(gp): Pass a test_list in fast, slow, ... instead of duplicating all the code CmTask #1571.
 @task
 def run_fast_tests(  # type: ignore
--- a/helpers/lib_tasks_pytest.py
+++ b/helpers/lib_tasks_pytest.py
@@ -364,11 +364,27 @@ def run_mock_tests(ctx):
     """
     Only run the 10 tests in helpers/test/test_lib_tasks_pytest.py that uses mock.
     """
+    hlitauti.report_task()
     test_path = "helpers/test/test_lib_tasks_pytest.py::Test_build_run_command_line1"
+    cmd = "pytest"
     for num in [1, 2, 5, 6, 7]:
         for version in ["t", "f"]:
-            ctx.run(f"pytest {test_path}::test_run_fast_tests{num}{version}")
-    return 0
+            cmd += f" {test_path}::test_run_fast_tests{num}{version} "
+    base_image = ""
+    # We need to add some " to pass the string as it is to the container.
+    #######cmd = f"'{cmd}'"
+    # We use "host" for the app container to allow access to the database
+    # exposing port 5432 on localhost (of the server), when running dind we
+    # need to switch back to bridge. See CmTask988.
+    extra_env_vars = ["NETWORK_MODE=bridge"]
+    docker_cmd_ = hlitadoc._get_docker_compose_cmd(
+    base_image, "dev", "", cmd, extra_env_vars=extra_env_vars
+    )
+    _LOG.info("cmd=%s", docker_cmd_)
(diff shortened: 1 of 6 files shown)


LibertasSpZ:
I also noticed that the error msg from the checkpoint is different from mine, presumably because the tests are run on real ck infrastructures:

  ```
  Test_build_run_command_line1.test_run_fast_tests1t ______________
  File "/app/helpers/test/test_lib_tasks_pytest.py", line 39, in test_run_fast_tests1t
    obj = hserver()
  TypeError: 'module' object is not callable
  ```

helpers/lib_tasks_pytest.py
```diff
@@ -359,6 +359,35 @@ def run_tests(  # type: ignore
     return rc
 
 
+@task
```
gpsaggese, line 362: Good job figuring this out.

The problem is that it's super specific for your case and so we can't check it in.

FYI
1) pytest has easy ways to execute a subset of tests (you can check the documentation)
2) if you *really* want to execute a subset of tests, you can create a testlist using markers. I would have been ok with just running all the class

IMO because this was just a tmp stuff for your debugging I would have created a bash one-liner

```
#!/bin/bash -xe
docker_cmd -c "pytest helpers/test/test_lib_tasks_pytest.py::Test_build_run_command_line1"
```

or just explicitly write the list of tests to run, and then run the throw away script.

We don't want to write code for tmp functionality.

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -12,29 +13,52 @@
...
+                tee_to_file,
+                n_threads,
+            )
+            exp = (
+                'pytest -m "not slow and not superslow" . '
+                "-o timeout_func_only=true --timeout 5 --reruns 2 "
+                '--only-rerun "Failed: Timeout" -n 1'
+            )
+            #timeout_in_sec = self._compute_timeout_depending_on_ck_infra()
+            #exp = exp.replace("--timeout 5", f"--timeout {timeout_in_sec}")
```
gpsaggese, line 55: Pls remove commented code. We don't want to leave traces of the history of the code in the code. Git does that for us

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -233,25 +314,33 @@ def test_run_fast_tests7(self) -> None:
...
+                n_threads,
+            )
+            exp = (
+                'pytest -m "not slow and not superslow" . '
+                "-o timeout_func_only=true --timeout 5 --reruns 2 "
+                '--only-rerun "Failed: Timeout" -n auto'
+            )
+            timeout_in_sec = 5 if mock_is_dev_ck else 50
+        #timeout_in_sec = self._compute_timeout_depending_on_ck_infra()
+            exp = exp.replace("--timeout 5", f"--timeout {timeout_in_sec}")
```
gpsaggese, line 335: IMO this is not worth it. Better to hardwire the two `exp` and passing it.

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -12,29 +13,52 @@
...
 
 _LOG = logging.getLogger(__name__)
 
 # pylint: disable=protected-access
 
-
 class Test_build_run_command_line1(hunitest.TestCase):
     
-    def _compute_timeout_depending_on_ck_infra(self) -> int:
+    def test_run_fast_tests1t(self) -> None:
```
gpsaggese, line 26: Long clear names are better than encoding

-> test_run_fast_tests1_on_ck_server

gpsaggese (commented):
Good work solving the problems, but the solution is more complicated / more difficult to maintain than what could be
It will take a bit of time to internalize it these criteria

helpers/lib_tasks_pytest.py
LibertasSpZ: Thanks @gpsaggese . This makes sense. 

To clarify the one-liner should be `invoke docker_cmd - c ...` though. Will try to find nicer solutions like this in future.

LibertasSpZ:
Thanks again for the comments @gpsaggese . I think this closes #518 and possibly #282 .

helpers/test/test_lib_tasks.py
```diff
@@ -131,6 +131,8 @@ def _cmd(cmd):
     hsystem.system(cmd)
 
 
+# TODO(ShaopengZ): fails on ck server. `gh auth login` issue.
```
gpsaggese, line 134: Ok, I think you mean it fails when running Sorrentum on CK server. This is "expected".
But cmamp on CK server should work.

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -1,163 +1,76 @@
...
 # pylint: disable=protected-access
 
+
 class Test_build_run_command_line1(hunitest.TestCase):
-    
-    def test_run_fast_tests1t(self) -> None:
+    def test_run_fast_tests1_on_ck_server(self) -> None:
         """
-        Basic run fast tests for is_dev_ck()=True.
+        Mock test for running fast tests on ck server.
```
gpsaggese, line 27: Nit: In comments let's use CK 

docs/Coding_Style_Guide.md:316:- Abbreviations

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -1,163 +1,76 @@
...
-            self.assert_equal(act, exp)
-    
-    # TODO(Shaopeng): factor out the common parts of the mocked tests if time allows.
-    def test_run_fast_tests1f(self) -> None:
+        exp = (
+            'pytest -m "not slow and not superslow" . '
+            "-o timeout_func_only=true --timeout 5 --reruns 2 "
+            '--only-rerun "Failed: Timeout" -n 1'
+        )
+        self._test_run_fast_tests1_helper(True, exp)
```
gpsaggese, line 34: Let's avoid "magic" values. Better to do:
```
is_dev_ck_return_value = True
self._test_run_fast_tests1_helper(is_dev_ck_return_value, exp)
```

Let's do this everywhere

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -214,25 +127,76 @@ class TestHello_World(hunitest.):
...
+
+    def test_run_fast_tests7_outside_ck_server(self) -> None:
+        exp = (
+            'pytest -m "not slow and not superslow" . '
+            "-o timeout_func_only=true --timeout 50 --reruns 2 "
+            '--only-rerun "Failed: Timeout" -n auto'
+        )
+        self._test_run_fast_tests7_helper(False, exp)
+
+    def _test_run_fast_tests1_helper(self, is_dev_ck_return_value, exp) -> None:
```
gpsaggese, line 180: For tests we allow a little more flexible naming based on increasing readability.

let's make the private functions `_test_run_fast_tests2_helper` public -> `run_fast_tests2_helper` (they can't start with `test` otherwise they are test methods), then we can move me before the caller tests and do something like

```
# #####################################

def run_fast_tests7

def test_run_fast_tests7_on_ck_server

def test_run_fast_tests7_outside_ck_server
```

so that it's easier to understand what tests does what

helpers/test/test_lib_tasks_pytest.py
```diff
@@ -214,25 +127,76 @@ class TestHello_World(hunitest.):
...
         """
-        Basic run fast tests tee-ing to a file. Mock depending on if mock_is_dev_ck is true.
-        
-        :param: mock_is_dev_ck: mocking the return_value of hserver.is_dev_ck.
+        Basic run fast tests.
+
+        :param: is_dev_ck_return_value: mocking the return_value of hserver.is_dev_ck.
```
gpsaggese, line 184: No need of `:` after param, use verbatim for vars, and function need ()

: param is_dev_ck_return_value: mocking the return value of `hserver.is_dev_ck()`

In general you can look at the rest of the code base to pick up the style

gpsaggese (commented):
Good work. Let's merge and then we can do another small follow up PR.

The goal is to focus / internalize the style, but I think you are picking it up.

helpers/test/test_lib_tasks_pytest.py
LibertasSpZ: Got it. thanks!

helpers/test/test_lib_tasks_pytest.py
LibertasSpZ: got it. Thanks.

Issue #512, opened 2023-08-03
Upload the local csv to google spreadsheet


yiyunlei:
- Interact with Google API to upload the local CSV to Google spreadsheet.
    - Use gspread_pandas
- After uploading to gsheet, delete the tmp file.

yiyunlei:
- ETA: Next Tuesday(2023-08-08)
  - Implement function and add to the notebook.

gpsaggese:
When do we use local files? Can we always read / modify gsheets without ever copying to local files?
Copying things around makes things more difficult, since you have multiple copies which creates confusion

The approach I suggest is to create a gsheet dir and then each step (notebook or PB) reads one gsheet as inoput and generates another one without overwriting.

Makes sense?

yiyunlei:
This is the second step to extract PB results to gsheet. #487 
(We use local files only in this case)
- I implemented the first step to get PB results from the PB server to local storage. #515 
- After downloading to local, we need to upload the result CSVs to gsheet so that we can use them as input for other steps.
- We don't have to copy these files manually, as it will automate the download of CSV files and upload to gheet after I implement these two steps.

gpsaggese:
Clear explanation. Now I understand.

Just make sure to download to local storage as tmp file, upload to gsheet, and delete the tmp file.
The idea is to avoid duplicated data

yiyunlei:
- ETA: Today(2023-08-09)
    - Finish up PR #525

PR #525, opened 2023-08-08
Sorr Task496 refine google sheets flow

yiyunlei:
Task #496 and #512


yiyunlei:
Ready to be merged.

linkedin/google_api/google_drive_api.py
```diff
@@ -20,23 +27,27 @@ def __init__(self):
...
         Create a new Google file (sheet or doc).
 
         :param gfile_type: str, the type of the Google file ('sheet' or 'doc').
         :param gfile_name: str, the name of the new Google file.
-        :param gdrive_folder: dict, the id and the name of the Google Drive folder.
+        :param gdrive_folder_id: the id of the Google Drive folder.
         :param user: str, the email address of the user to share the Google file (Optional).
-        :return: None
         """
         try:
```
gpsaggese, line 52: I would not do try-except. If there is an error I don't see how we can recover from it and we just need to stop

linkedin/google_api/google_drive_api.py
```diff
@@ -49,28 +60,50 @@ def create_empty_google_file(
...
+            _LOG.error(err)
+
+    def create_google_drive_folder(self, folder_name: str, parent_folder_id: str) -> str:
+        """
+        Create a new Google Drive folder inside the given folder.
+
+        :param folder_name: str, the name of the new Google Drive folder.
+        :param parent_folder_id: str, the id of the parent folder.
+        """
+        try:
```
gpsaggese, line 82: same here

linkedin/notebooks/create_gsheets_and_fetch_result_csv.py
```diff
@@ -0,0 +1,145 @@
+# ---
```
gpsaggese, line 1: Not sure I understand the goal of this notebook.

In my mental model you do multiple steps:
1) run phantom to get the results search
2) notebook to process the search data (read from Phantom, save in gsheet, read from gsheet, filter, save to gsheet)
3) run phantom to get the profile
4) notebook to process the profile data (read from Phantom, save in gsheet, read from gsheet, filter, save to gsheet)

The fact that we have already both phases 1) and 3) done is an artifact. The real flow won't have that, since we run the phases serially.

We want to have only 2 notebooks.
Each notebook should be "idempotent": if the data has already been extracted from PB then skip.

Does it make sense?

gpsaggese (commented):
The code is well done.

I think there is a bit of confusion on the flow. See comments below.

Let's merge and then follow up

linkedin/notebooks/create_gsheets_and_fetch_result_csv.py
yiyunlei: Yeah, I understood what you were saying. I should put the step of retrieving data from PB in the two processing notebooks.

This new notebook is meant to create empty folders and gsheets. I put them in this new notebook so that we can create empty folders and gsheets together and use them in the processing notebooks.

I can also put create folder/gsheets functions in the two processing notebooks, but the code will be a bit duplicated.

linkedin/notebooks/create_gsheets_and_fetch_result_csv.py
```diff
@@ -0,0 +1,145 @@
...
+# jupyter:
+#   jupytext:
+#     text_representation:
```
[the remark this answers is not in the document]
yiyunlei: Agree, I will describe the new flow in the gdoc and request your review before I implement.

linkedin/notebooks/create_gsheets_and_fetch_result_csv.py
gpsaggese: > I should put the step of retrieving data from PB in the two processing notebooks.

Correct. You want to create dirs and files as you go and not ahead of time. In this way, you can make the code "idempotent" and "incremental"

See some notes
docs/Jupyter_notebook_best_practices.md:29:  * [Make cells idempotent](#make-cells-idempotent)

> I can also put create folder/gsheets functions in the two processing notebooks, but the code will be a bit duplicated.

If you keep the API and the functions in a lib then the code in the notebook doesn't have repetition, but it looks like we do when creating dirs on local disk

```
hio.create_gdrive_dir(path, incremental=True)
```

We want to mimic the Google API with functions in hio (e.g., `does_file_exists`, `create_dir`)

```
if not incremental or not does_file_exists(file):
    # download from PB and create a gsheet in the dir
```

yiyunlei: Got it, thanks for explaining.

Issue #529, opened 2023-08-09
Generalize the switch to take a marker


LibertasSpZ:
Per discussion in the [google document](https://docs.google.com/document/d/1Qm54LwNRlBwzYroDGviW32NevK_V8L4Sufsg5u6IKwI/edit) on Sorr Task #283 , make the switches in `run_fast_tests` accept strings of test names.

PR #530, opened 2023-08-10
 SorrTask529_Generalize_the_switch_to_take_a_marker 

LibertasSpZ:
[1] Finish issue #529 (and the google doc cited there).
[2] Fix/skip all fast tests hanging outside CK infra and all fast tests that requires aws.
  - As of now, `i run_fast_tests` works on my laptop with no hang, error, or failure, taking about 14 min (note the extended timeout for fast tests outside CK infra), even without using switches. Please double check. If this checks out, #282 should be closed.
  
Sorry I didn't do a smaller PR because to test the switches I had to cure all the hanging fast tests.

Note: the checkpoint on Sorrentum skips the "Runs only on CK infra" tests. Maybe someone can check those tests on CK infra. The command would be `i run_fast_tests -r requires_ck_infra`.


helpers/test/test_git.py
```diff
@@ -42,6 +43,11 @@ def test_is_inside_submodule1(self) -> None:
         func_call = "hgit.is_inside_submodule()"
         _execute_func_call(func_call)
 
+    @pytest.mark.requires_ck_infra #outside CK infra hgit.is_amp() usually hangs
```
gpsaggese, line 46: We put comments on separate lines and we leave a space, start with capital letter and end with period. I think the linter does a lot of this for you. Have you set it up?

```
# Outside CK infra hgit.is_amp() usually hangs, so we skip it.
@pytest.mark.requires_ck_infra
```

Also in this case it's is_dev_ck and not is_amp in the comment I believe.

Let's be super careful at the details and re-read the PR with fresh eyes.

helpers/test/test_git.py
```diff
@@ -171,6 +177,11 @@ def test_get_repo_name4(self) -> None:
         exp = "dev_tools"
         self.assert_equal(act, exp)
 
+    @pytest.mark.requires_ck_infra
+    @pytest.mark.skipif(
+        not hserver.is_dev_ck(), #outside CK infra, hangs at hgit.is_in_amp_as_supermodule()
```
gpsaggese, line 182: Same

helpers/test/test_git.py
```diff
@@ -182,6 +193,11 @@ def test_get_all_repo_names1(self) -> None:
         exp = ["amp", "cmamp", "dev_tools"]
         self.assert_equal(str(act), str(exp))
 
+    @pytest.mark.requires_ck_infra
+    @pytest.mark.skipif(
```
gpsaggese, line 197: I'm confused. Why do we need this 

```
@pytest.mark.skipif(
        not hserver.is_dev_ck(), #outside CK infra, hangs at hgit.is_in_amp_as_supermodule()
        reason = "Run only on CK infra"
    )
```
if the test is marked as requires_ck_infra?

If we are outside CK infra then is_dev_ck is false so the two conditions are equivalent?

helpers/test/test_git.py
LibertasSpZ: Will run linter in the next PR. 
I believe the comment should say `hgit.is_amp()` because we are skipping the function because it calls `hgit.is_amp()` and hangs outside CK. `is_dev_ck()` does not hang.

helpers/test/test_git.py
LibertasSpZ: See the above.

helpers/test/test_git.py
LibertasSpZ: This is for `i run_fast_tests` to pass even without using the switches.

If we cancel these skips then `i run_fast_tests` hangs / fails etc when ran outside CK, and user has to run `i run_fast_tests -k requires_ck_infra`.

If the latter is intended [that is, user cannot run `i run_fast_tests` on laptops and pass all tests unless they turn on the skip switch] then we can cancel these conditional skips.

helpers/lib_tasks_pytest.py
```diff
@@ -366,45 +366,57 @@ def run_fast_tests(  # type: ignore
     stage="dev",
     version="",
     pytest_opts="",
+    run_only_test_list_as_string="",
```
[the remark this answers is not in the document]
LibertasSpZ: Can do, but then we need a new name for the list formed by splitting the string. I think if we do
`run_only_test_list = run_only_test_list.split(',')` we would get a type mismatch error.

helpers/test/test_git.py
LibertasSpZ: I think what we can do is: delete the `pytest.mark.skipif( OUTSIDE_CK_INFRA) ` and add the following to the doc string of `run_fast_tests`:

  ```
  For users running outside CK infra, to pass all fast tests, please invoke with the -k requires_ck_infra switch.
  ```

helpers/lib_tasks_pytest.py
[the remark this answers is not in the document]
gpsaggese: You can resolve (after you made the change) the comments that have been addressed (like this one)

helpers/test/test_git.py
gpsaggese: Personally I wouldn't try to make the comment dependent on the code if you put it close to the problem.

```
# Outside CK infra, the following call hangs, so we skip it.
```

helpers/test/test_git.py
gpsaggese: If one uses `pytest` then he / she needs to use the right markers. If you want to automatically use the markers then use `i run_fast_tests`. In fact `i run_fast_tests` should detect `not is_dev_ck()` and add the marker `requires_ck_infra` automatically.

Does it make sense?

LibertasSpZ: Yes. We had a similar approach in some earlier version and I will put the feature back in the next PR. Thanks for clearing this up.

helpers/lib_tasks_pytest.py
```diff
@@ -366,45 +366,73 @@ def run_fast_tests(  # type: ignore
...
+    test_list_name = "fast_tests"
+    # If we are running outside the CK server, tests that requires CK infra
+    # should be automatically skipped.
+    is_outside_ck_infra = not hserver.is_dev_ck()
+    if is_outside_ck_infra:
+        if skip_test_list:
+            skip_test_list = "requires_ck_infra," + skip_test_list
+        else:
+            skip_test_list = "requires_ck_infra"
+    # split the run_only_test_list and skip_test_list strings into lists.
```
gpsaggese, line 410: -> Split the `run_only_test_list` and `skip...`

Each comment is an English phrase starting with capital, ending with `.` and using verbatim when you use a variable.
Can you follow this everywhere?

helpers/lib_tasks_pytest.py
```diff
@@ -366,45 +366,73 @@ def run_fast_tests(  # type: ignore
...
+    else:
+        run_only_test_list = []
+    if skip_test_list:
+        # This works as expected when there is a single test in the list.
+        skip_test_list = skip_test_list.split(",")
+        _LOG.warning(f"Skipping the tests inside: {skip_test_list}")
+    else:
+        # when running inside CK infra, it is still possible for skip_test_list to be empty.
+        skip_test_list = []
+    # form marker strings for pytest -m using " and " and " not ".
```
gpsaggese, line 424: -> Convert marker strings for `pytest -m` using `and` and `not`.

helpers/lib_tasks_pytest.py
```diff
@@ -366,45 +366,73 @@ def run_fast_tests(  # type: ignore
...
+        run_only_test_list = run_only_test_list.split(",")
+        _LOG.warning(f"Running only tests inside: {run_only_test_list}")
+    else:
+        run_only_test_list = []
+    if skip_test_list:
+        # This works as expected when there is a single test in the list.
+        skip_test_list = skip_test_list.split(",")
+        _LOG.warning(f"Skipping the tests inside: {skip_test_list}")
+    else:
+        # when running inside CK infra, it is still possible for skip_test_list to be empty.
```
gpsaggese, line 422: -> When ...

helpers/lib_tasks_pytest.py
```diff
@@ -366,45 +366,73 @@ def run_fast_tests(  # type: ignore
...
     :param n_threads: the number of threads to run the tests with
         - "auto": distribute the tests across all the available CPUs
     :param git_clean_: run `invoke git_clean --fix-perms` before running the tests
-    :param only_requires_ck_infra_tests: select ONLY ck infra related tests.
-    :param skip_requires_ck_infra_tests: skip ck infra related tests.
     :param kwargs: kwargs for `ctx.run`
     """
     hlitauti.report_task()
-    test_list_name = "fast_tests"
     hdbg.dassert(
```
gpsaggese, line 397: At this point we should factor out this function and then unit test carefully. 
Don't worry for now. We'll have an intern taking care of it

gpsaggese (commented):
Minor tweaks and then merge.

Good job.

Issue #481, opened 2023-07-31
Unit test `plot_cols()`


samarth9008:
We need to unit test `plot_cols()` from [core/plotting/misc_plotting.py](https://github.com/sorrentum/sorrentum/blob/ecf0a5efc7c5cabd5a7d36be258ebb3851c415dc/core/plotting/misc_plotting.py#L372)

Unit testing of a plotting function consists of 2 parts:
- Call the function in a `core/notebooks/gallery_signal_processing.ipynb` to check that it works (do not commit, just confim how it looks in the PR with a screenshot)
- Add a smoke test for the function to `core/plotting/test/test_plots.py`. Use the same inputs as in the `core/notebooks/gallery_signal_processing.ipynb`

For reference see https://github.com/sorrentum/sorrentum/pull/393

FYI @DanilYachmenev

samarth9008:
@DanilYachmenev Can you PTAL bcz the doc says `Already in a gallery but not covered`?

samarth9008:
Whats the update @brandonnsc

PR #543, opened 2023-08-16
SorrTask 481 Unit test plot_cols()

zli00185:
#481


zli00185:
The function and data was confirm workable. Here is the screenshot.
Screenshot 2023-08-16 at 8 47 33 PM

core/plotting/test/test_plots.py
```diff
@@ -29,6 +29,20 @@ def get_plot_spectrum1() -> pd.DataFrame:
         df = pd.DataFrame(data)
         return df
 
+    @staticmethod
+    def get_plot_col1() -> pd.DataFrame:
```
DanilYachmenev, line 33: - `get_plot_col1()` and `get_plot_effective_correlation_rank1()` are identaica
- seems like `get_plot_heatmap1()` and `get_plot_spectrum1()` can be easily replaced by  `get_plot_col1()` as well

So my proposal is
- create `get_test_plot_df()` which is a copy of this func
- deprecate all the functions mentioned above and replace their usage with `get_test_plot_df()`
- rerun test and notebook, make sure everything passes and looks good

core/plotting/test/test_plots.py
```diff
@@ -170,3 +184,10 @@ def test_plot_spectrum2(self) -> None:
         _, axes = plt.subplots(2, 2, figsize=(20, 10))
         axes_flat = axes.flatten()
         cplmiplo.plot_spectrum(signal=test_df, axes=axes_flat)
+
+    def test_plot_cols(self) -> None:
```
DanilYachmenev, line 188: -> `test_plot_cols1`

DanilYachmenev (commented):
lg, see the further suggestions

core/plotting/test/test_plots.py
zli00185: Should I make the function public?

zli00185:
All unit test pass
Screenshot 2023-08-17 at 11 01 22 AM

core/plotting/test/test_plots.py
```diff
@@ -18,31 +18,20 @@ class Test_plots(unittest.TestCase):
     """
 
     @staticmethod
-    def get_plot_spectrum1() -> pd.DataFrame:
+    def get_test_plot_df() -> pd.DataFrame:
```
DanilYachmenev, line 21: for all test methods we finish their names with a respective integers
so here it should be `get_test_plot_df1`

core/plotting/test/test_plots.py
```diff
@@ -18,31 +18,20 @@ class Test_plots(unittest.TestCase):
...
         """
-        Generate a test DataFrame for the plot_spectrum function.
+        Generate a test DataFrame for the plot_cols, plot_spectrum,
+        plot_effective_correlation_rank function.
```
DanilYachmenev, line 24: `Generate a DataFrame with random values to test plotting functions.`

core/plotting/test/test_plots.py
DanilYachmenev, line 21: replace `get_plot_heatmap1()` with this one as well

DanilYachmenev, line 21: And now lets also make replacements in the gallery notebook
you deprecate the functions so make sure you update all their calls

check that test for the gallery notebook passes

zli00185:
Not sure what cause the test fail. I ran the purest and all passed

DanilYachmenev (commented):
lgtm

test failures are not related to the PR changes

Issue #546, opened 2023-08-21
Test `helpers/hpandas.py` functions with multiindex dataframe


DanilYachmenev:
Now `dassert_index_is_datetime()` and `dassert_time_indexed_df()` from `helpers/hpandas.py` can be used with multiindex dataframes

Add a simple unit tests with multiindex dataframes for the mentioned functions to `helpers/test/test_hpandas.py`
Read more about how we do unit tests in the corresponding [doc](https://github.com/sorrentum/sorrentum/blob/master/docs/Unit_tests.md)

Also make sure to take a look at other unit tests in the file and try to comply with the format of already existing tests

FYI @tamriq

PR #548, opened 2023-08-21
Final Changes to the functions #546

Sameep2808:
@DanilYachmenev Task #546. Hey I have created a pull request please review. I also noticed that when I run `i run_fast_tests` there are multiple files getting generated I have removed them from the push right now but do let me know if they are supposed to be pushed


helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,284 @@ def test3(self) -> None:
...
         """
         self.assert_equal(act, exp, fuzzy_match=True)
+
+
+# #############################################################################
+
+
+class Test_multiindex_df_datetime(hunitest.TestCase):
+    @staticmethod
+    def get_multiindex_df_with_datetime_index(test_type: int) -> pd.DataFrame:
```
DanilYachmenev, line 3243: well, its an overkill)
very extensive testing
which does the job, but here we have no need to check for so many different cases
and no need to test it on increasing or decreasing indices

the assert checks if index is timestamp type - this is all we need to check here
so what I'd do is to
- create a `helper()` method that will return a multiindex df to test the function on
- 1 test method to test if function passes on it
- 1 test method with edited `helper()` df that should not pass the assert (like add 1 `None` to it)
- convert multiindex df to a simple df (like `df[some_index]` from the original multiindex df) and do the same

so lets start  with just 4 tests
2 for multi: 1 pass, 1 fail
2 for simple index: 1 pass 1 fail

also, lets make test df generation as simple as possible, like in `get_multiindex_df_with_non_datetime_index`
so you can create a datetime index with `pd.date_range()` and fill rows with random values or `range()` func

in short - simplify
it good that it works but we also should strive to make the code as simple as possible so we do not bloat the tech debt and reduce time on unnecessary checks and over-optimization

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,284 @@ def test3(self) -> None:
...
+        self.assert_equal(act, exp, fuzzy_match=True)
+
+    def test4(self) -> None:
+        """
+        Test dataframe time indexed and well-formed.
+        """
+        df = self.get_multiindex_df_with_datetime_index(0)
+        allow_empty = False
+        strictly_increasing = False
+        hpandas.dassert_time_indexed_df(df, allow_empty, strictly_increasing)
```
DanilYachmenev, line 3385: like here - checking for strictly increasing index is not a responsibility of this assert but of a nested one, so no need for it here

DanilYachmenev (commented):
its a good work, I see your attention to details

However, we need to simplify your code here
Nothing wrong here, but overkilling is not always needed and if so - it just steals the time from both collaborator and reviewer

so yeah, see the suggestions below

helpers/test/outcomes/TestCheckDataFrame1.test_check_df_not_equal3/output/test_df.txt
```diff
@@ -1 +1 @@
-abc
+abc
```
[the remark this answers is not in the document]
Sameep2808: Okay Ill start using `amp/dev_scripts/clean_up_text_files.sh` before PR.

Yes I know I was using `> pytest helpers/test/test_hpandas.py:: Test_multiindex_df_datetime` all the time just ran all tests once before committing to crosscheck if everything was working.

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,284 @@ def test3(self) -> None:
...
         self.assert_equal(act, exp, fuzzy_match=True)
+
+
+# #############################################################################
+
+
+class Test_multiindex_df_datetime(hunitest.TestCase):
+    @staticmethod
+    def get_multiindex_df_with_datetime_index(test_type: int) -> pd.DataFrame:
+        if test_type == 0:  # Strictly Increasing
```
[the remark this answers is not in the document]
Sameep2808: I'll keep this in mind thanks for informing

helpers/test/test_hpandas.py
Sameep2808: Yes even I thought of using random for df generation but then referred the code above and thought maybe you guys want to keep track of exact input for the test cases hence implemented static df but now have updated it please take a look.

Thanks for the detailed review will keep it simpler next time rather than extensive testing

DanilYachmenev (commented):
good progress but needs to be yet simpler and clearer

as always, before every commit run linter + run the tests to check if they pass

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,166 @@ def test3(self) -> None:
...
+            ]
+        else:
+            datetime_index = ["string1", "string2"]
+        iterables = [
+            ["column1", "column2"],
+            ["subcolumn1", "subcolumn2", "subcolumn3", "subcolumn4"],
+        ]
+        index = pd.MultiIndex.from_product(iterables, names=[None, "timestamp"])
+        nums = np.random.uniform(-2, 2, size=(2, 8))
+        df = pd.DataFrame(nums, index=datetime_index, columns=index)
```
DanilYachmenev, line 3263: well, I expect we need to pass `index` as `index`
and check if assert passes if `pd.MultiIndex` contains datetime values

Снимок экрана 2023-08-23 в 18 13 12

now you pass just 1-level index and multi-level column

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,166 @@ def test3(self) -> None:
...
+        """
+        self.assert_equal(act, exp, fuzzy_match=True)
+
+
+# #############################################################################
+
+
+class Test_multiindex_df_timezone(hunitest.TestCase):
+    @staticmethod
+    def get_multiindex_df_with_datetime_index(
```
DanilYachmenev, line 3309: -> `helper()`

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,166 @@ def test3(self) -> None:
...
+        timezone: bool,
```
DanilYachmenev, line 3310: why `timezone` here? I'd expect this helper to be the same as in the previous test

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,166 @@ def test3(self) -> None:
...
+
+        datetime_index = pd.date_range(start=start_time, end=end_time, freq="T")
+
+        iterables = [
+            ["column1", "column2"],
+            ["subcolumn1", "subcolumn2", "subcolumn3", "subcolumn4"],
+        ]
+        index = pd.MultiIndex.from_product(iterables, names=[None, "timestamp"])
+        nums = np.random.uniform(-2, 2, size=(5, 8))
+        df = pd.DataFrame(nums, index=datetime_index, columns=index)
```
DanilYachmenev, line 3328: same here - make multiindex index

and never name `columns=index`
its just misleading
`columns=columns`, `index=index`

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,166 @@ def test3(self) -> None:
             rows=2 ['M', 'N']
         """
         self.assert_equal(act, exp, fuzzy_match=True)
+
+
+# #############################################################################
+
+
+class Test_multiindex_df_datetime(hunitest.TestCase):
```
DanilYachmenev, line 3241: -> `Test_dassert_index_is_datetime`

name test classes after the tests func names
same below

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,166 @@ def test3(self) -> None:
...
+        simple_df.columns = ["column1", "column2", "column3", "column4"]
+        allow_empty = False
+        strictly_increasing = False
+        hpandas.dassert_time_indexed_df(
+            simple_df, allow_empty, strictly_increasing
+        )
+
+    def test4(self) -> None:
+        """
+        Test dataframe for index not containing timezone.
```
DanilYachmenev, line 3376: unit test the same cases as in the previous test

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,166 @@ def test3(self) -> None:
...
+        """
+        timezone = False
+        df = self.get_multiindex_df_with_datetime_index(timezone)
+        simple_df = df.stack(level=0).reset_index(level=1, drop=True)
+        simple_df.index.name = "timestamp"
+        simple_df.columns = ["column1", "column2", "column3", "column4"]
+        allow_empty = False
+        strictly_increasing = False
```
DanilYachmenev, line 3384: too complicated and IMO unnecessary in this case
just use the input df
if you want to move from multi to single index, well, just select a level in it
for this you'd have to design a test df in a way that will give you desired inputs
I'd suggest you to make the outer level index as string and inner level index as datetime
so smth like
```
                                       col1   col2 
index1          2022-02-22 00:00:01      12     10
                2022-02-22 00:00:02      10      8
index2          2022-02-22 00:00:01       ...
```

and then you can always do `df["index1"]`

helpers/test/test_hpandas.py
Sameep2808: @DanilYachmenev where is this code from Im unable to find it the version I have is:
```
def dassert_index_is_datetime(
    obj: Union[pd.Index, pd.DataFrame, pd.Series],
    msg: Optional[str] = None,
    *args: Any,
) -> None:
    """
    Ensure that the dataframe has an index containing datetimes.
    """
    index = _get_index(obj)
    hdbg.dassert_isinstance(index, pd.DatetimeIndex, msg, *args)
```

helpers/test/test_hpandas.py
Sameep2808: Hey Dan I have implemented the helper function as you told but then I doubt the `dassert_index_is_datetime()` function can handle Multi index as it needs as I dont have below version of code is the master updates??

> Снимок экрана 2023-08-23 в 18 13 12

The function I have:
```
def dassert_index_is_datetime(
    obj: Union[pd.Index, pd.DataFrame, pd.Series],
    msg: Optional[str] = None,
    *args: Any,
) -> None:
    """
    Ensure that the dataframe has an index containing datetimes.
    """
    index = _get_index(obj)
    hdbg.dassert_isinstance(index, pd.DatetimeIndex, msg, *args)
```

helpers/test/test_hpandas.py
Sameep2808: The earlier one was index with datetime this one is index without date time for single dataframe

helpers/test/test_hpandas.py
DanilYachmenev: Oh, I see
It is absent because this part of the code was not integrated to `sorrentum` yet

in order to not let your work go in vain, lets do this small integration ourselves
replace the corresponding functions code with the following:
```
def dassert_index_is_datetime(
    obj: Union[pd.Index, pd.DataFrame, pd.Series],
    msg: Optional[str] = None,
    *args: Any,
) -> None:
    """
    Ensure that the dataframe has an index containing datetimes.

    Works for both single and multi-indexed dataframes.
    """
    index = _get_index(obj)
    if isinstance(index, pd.MultiIndex):
        # In case of multi index check that at least one level is a datetime.
        is_any_datetime = any(
            isinstance(level, pd.DatetimeIndex) for level in index.levels
        )
        hdbg.dassert(is_any_datetime, msg, *args)
    else:
        hdbg.dassert_isinstance(index, pd.DatetimeIndex, msg, *args)


def dassert_time_indexed_df(
    df: pd.DataFrame, allow_empty: bool, strictly_increasing: bool
) -> None:
    """
    Validate that input dataframe is time indexed and well-formed.

    Works for both single and multi-indexed dataframes.

    :param df: dataframe to validate
    :param allow_empty: allow empty data frames
    :param strictly_increasing: if True the index needs to be strictly increasing,
      instead of just increasing
    """
    # Verify that Pandas dataframe is passed as input.
    hdbg.dassert_isinstance(df, pd.DataFrame)
    if not allow_empty:
        # Verify that a non-empty dataframe is passed as input.
        hdbg.dassert_lt(0, df.shape[0])
        # Verify that the dataframe has at least 1 column.
        hdbg.dassert_lte(1, len(df.columns))
    # Verify that the index is increasing.
    if strictly_increasing:
        dassert_strictly_increasing_index(df)
    else:
        dassert_increasing_index(df)
    # Check that the index is in datetime format.
    dassert_index_is_datetime(df)
    # Check that the passed timestamp has timezone info.
    index_item = df.index[0]
    if isinstance(index_item, tuple):
        # In case of multi index assume that the first level is a datetime.
        index_item = index_item[0]
    hdateti.dassert_has_tz(index_item)
```

helpers/hpandas.py
```diff
@@ -219,7 +230,11 @@ def dassert_time_indexed_df(
     # Check that the index is in datetime format.
     dassert_index_is_datetime(df)
     # Check that the passed timestamp has timezone info.
-    hdateti.dassert_has_tz(df.index[0])
+    index_item = df.index[0]
+    if isinstance(index_item, tuple):
+        # In case of multi index assume that the first level is a datetime.
+        index_item = index_item[0]
+    hdateti.dassert_has_tz(index_item)
```
DanilYachmenev, line 237: here it seems that the assert will pass only if `index_outer` is datetime that has a timezone
therefore, in `helper` make `index_outer` to comply with it

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,210 @@ def test3(self) -> None:
...
+                pd.Timestamp("2022-01-01 21:10:00", tz="UTC"),
+            ]
+        else:
+            datetime_index = ["string1", "string2"]
+        iterables = [
+            ["column1", "column2"],
+            ["subcolumn1", "subcolumn2", "subcolumn3", "subcolumn4"],
+        ]
+        iterables2 = [["index1", "index2"], datetime_index]
+        index = pd.MultiIndex.from_product(
```
DanilYachmenev, line 3263: set `index_outer` before
and `index` should be derived from `index_outer` and `index_inner`

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,210 @@ def test3(self) -> None:
...
+            nums = np.random.uniform(-2, 2, size=(4, 8))
+            df = pd.DataFrame(nums, index=index, columns=columns)
+        else:
+            nums = np.random.uniform(-2, 2, size=(2, 8))
+            df = pd.DataFrame(nums, index=datetime_index, columns=columns)
+        return df
+
+    def test1(self) -> None:
+        """
+        Test dataframe for index containing datetimes.
```
DanilYachmenev, line 3278: -> `Check that dataframe index is datetime type.`
other docstrings in a similar way

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,210 @@ def test3(self) -> None:
...
+        hpandas.dassert_time_indexed_df(df, allow_empty, strictly_increasing)
+
+    def test2(self) -> None:
+        """
+        Test multiindex dataframe for index 0 not containing datetimes.
+        """
+        index_is_datetime = False
+        df = self.helper(index_is_datetime)
+        allow_empty = False
+        strictly_increasing = False
```
DanilYachmenev, line 3378: since `allow_empty` and `strictly_increasing` are also this func params, we need the following test cases:
- `allow_empty=False, strictly_increasing=True`, `df` is multiindex, test should pass
- `allow_empty=False, strictly_increasing=False`, `df` is multiindex, but 1 datetime index value is replaced, so the assert fails because `strictly_increasing` cond is not satisfied, test should fail
- `allow_empty=True, strictly_increasing=False`, `df` is empty, test should pass
- `allow_empty=False, strictly_increasing=False`, `df` is single index, but not datetime, test should fail

Sameep2808: @DanilYachmenev Yes I had done this but deleted cause you said the tests were extensive. No problem will add them back.
When I had written this test I noticed that it used to fail at `dassert_index_is_datetime(df)` indirectly not allowing for empty dataframe.

> * `allow_empty=True, strictly_increasing=False`, `df` is empty, test should pass

DanilYachmenev: good that you encountered it in the first place
however, here's an example of how lots of code can distract from the details

see, the first review is usually a high-level one
I cant spend my time on checking lots of code if I see that most of it is not needed, which was the case in the beginning
so from here you have 2 options:
- make a draft PR with just interfaces of the functions that you're going to make with short description of what they will contain. After agreed, write a code. e.g.
```
class Test_...
    def helper() -> pd.DataFrame
        """
        Create a dataframe with test data
        """
        
        def test1()
            """
            Check the case when ...
            """

        def test2()
            """
            Check the case when ...
            """
``` 
- write short self-explanatory code. usually, several lines each method and a short docstring

I opted for the 2nd approach here since these funcs are not of a highest priority
but as you can see, backs and forth happen in such cases

if you want to avoid it further on you can use draft approach

remember the principle of a collaborator: your major goal is to save time for your reviewer, not only implement what is asked in the issue.
After all, the reviewer is the one who hold the responsibility for the code and when evaluating the collaborator, one of the most factors for reviewers is how time-effective is he/she for the team

Sameep2808:
@DanilYachmenev I tried to avoid adding params in the helper function but then had to in-order to cover all the test cases suggested do let me know if you have any suggestions.

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+
+
+# #############################################################################
+
+
+class Test_dassert_index_is_datetime(hunitest.TestCase):
+    @staticmethod
+    def helper(
+        index_is_datetime: bool,
+    ) -> pd.DataFrame:
```
DanilYachmenev, line 3245: can be fit in 1 line
`def helper(index_is_datetime: bool) -> pd.DataFrame:`

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+    def helper(
+        index_is_datetime: bool = True,
+        switch_indexs: bool = False,
+        strictly_increasing: bool = True,
+    ) -> pd.DataFrame:
+        """
+        Get simple multi-index dataframe to unit test the assert.
+        """
+        if index_is_datetime:
+            # Get datatime index in dataframe.
```
DanilYachmenev, line 3328: remove, not needed

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+            if strictly_increasing:
+                # Datetime index is strictly increasing.
```
DanilYachmenev, line 3330: remove this one too

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+                    pd.Timestamp("2022-01-01 21:00:00", tz="UTC"),
+                    pd.Timestamp("2022-01-01 21:00:00", tz="UTC"),
+                ]
+        else:
+            index_outer = ["string1", "string2"]
+        index_inner = ["index1", "index2"]
+        if switch_indexs:
+            iterables = [index_inner, index_outer]
+        else:
+            iterables = [index_outer, index_inner]
```
DanilYachmenev, line 3346: no need for this switch
lets just make outer level datetime type and test only the 4th case from here https://github.com/sorrentum/sorrentum/pull/548/files#r1304728881

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+        allow_empty = False
+        strictly_increasing = False
+        hpandas.dassert_time_indexed_df(df, allow_empty, strictly_increasing)
+
+    def test2(self) -> None:
+        """
+        Check that multi-index dataframe doesn't have datetime type at index 0.
+        """
+        df = self.helper(index_is_datetime=False)
+        print(df.index[0][0])
```
DanilYachmenev, line 3367: no need

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+        with self.assertRaises(AssertionError) as cm:
+            hpandas.dassert_time_indexed_df(df, allow_empty, strictly_increasing)
+        act = str(cm.exception)
+        exp = r"""
+        * Failed assertion *
+        cond=False
+        """
+        self.assert_equal(act, exp, fuzzy_match=True)
+
+    def test3(self) -> None:
```
DanilYachmenev, line 3379: so no need for this one

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+
+# #############################################################################
+
+
+class Test_dassert_time_indexed_df(hunitest.TestCase):
+    @staticmethod
+    def helper(
+        index_is_datetime: bool = True,
+        switch_indexs: bool = False,
+        strictly_increasing: bool = True,
```
DanilYachmenev, line 3322: do not use default values for helpers in tests
specify params explicitly in each test so the code if self explanatory and does not require to get back to `helper()` in order to check for param values

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+            iterables = [index_outer, index_inner]
+        index = pd.MultiIndex.from_product(iterables, names=["name1", "name2"])
+        columns = ["column1", "column2"]
+        nums = np.random.uniform(-2, 2, size=(4, 2))
+        df = pd.DataFrame(nums, index=index, columns=columns)
+        return df
+
+    def test1(self) -> None:
+        """
+        Check that multi-index dataframe has datetime type at index 0.
```
DanilYachmenev, line 3355: -> `Check that multi-index df has a datetime outer index level.`

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+        """
+        df = self.helper()
+        allow_empty = False
+        strictly_increasing = False
+        hpandas.dassert_time_indexed_df(df, allow_empty, strictly_increasing)
+
+    def test2(self) -> None:
+        """
+        Check that multi-index dataframe doesn't have datetime type at index 0.
```
DanilYachmenev, line 3364: -> `Check that an assert is raised if multi-index df outer index level is datetime.`

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+        simple_df = df.loc["index1"]
+        allow_empty = False
+        strictly_increasing = False
+        hpandas.dassert_time_indexed_df(
+            simple_df, allow_empty, strictly_increasing
+        )
+
+    def test4(self) -> None:
+        """
+        Check that single-index dataframe doesn't have datetime type at index
```
DanilYachmenev, line 3393: -> `Check that an assert is raised if df index is not datetime.`

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+                simple_df, allow_empty, strictly_increasing
+            )
+        act = str(cm.exception)
+        exp = r"""
+        * Failed assertion *
+        Instance of 'Index(['string1', 'string2'], dtype='object', name='name2')' is '<class 'pandas.core.indexes.base.Index'>' instead of '<class 'pandas.core.indexes.datetimes.DatetimeIndex'>'
+        """
+        self.assert_equal(act, exp, fuzzy_match=True)
+
+    def test5(self) -> None:
```
DanilYachmenev, line 3411: no need, basically it is +- the same as `test1`

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+        """
+        df = self.helper()
+        allow_empty = False
+        strictly_increasing = True
+        hpandas.dassert_time_indexed_df(df, allow_empty, strictly_increasing)
+
+    def test6(self) -> None:
+        """
+        Check that multi-index dataframe has datetime type at index 0 and is
+        not strictly increasing.
```
DanilYachmenev, line 3424: -> `Check that an assert is raised if multi-index df datetime index is not strictly increasing.`

adjust the length to 80 chars everywhere

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,214 @@ def test3(self) -> None:
...
+        exp = r"""
+        * Failed assertion *
+        cond=False
+        Duplicated rows are:
+                                            column1   column2
+        name1                     name2
+        2022-01-01 21:00:00+00:00 index1    -0.122140 -1.949431
+                                index2     1.303778 -0.288235
+                                index1     1.237079  1.168012
+                                index2     1.333692  1.708455
```
DanilYachmenev, line 3441: tweak the test df so the assert is different
here the failure is rather about duplicating indices
so for `strictly_increasing=False` case make a df which df index is decreasing

Sameep2808: So I made the index same because there are 2 increasing cases 
1. Just Increasing: It allows duplication of indices but stops for decreasing indices.
2. Strictly Increasing: It stops both duplication of indices as well as decreasing indices

Sameep2808: @DanilYachmenev Just found out that `dassert_increasing_index()` cannot handle multiindex dataframe it fails at [this line](https://github.com/sorrentum/sorrentum/blob/52821f89bc16fecb4e0f67aca12f7e5c7baa8913/helpers/hpandas.py#L146) with the following error:

```
______________________ Test_dassert_time_indexed_df.test4 ______________________
  File "/app/helpers/test/test_hpandas.py", line 3396, in test4
    hpandas.dassert_time_indexed_df(df, allow_empty, strictly_increasing)
  File "/app/helpers/hpandas.py", line 229, in dassert_time_indexed_df
    dassert_increasing_index(df)
  File "/app/helpers/hpandas.py", line 155, in dassert_increasing_index
    mask = np.diff(index) <= pd.Timedelta(seconds=0)
  File "<__array_function__ internals>", line 180, in diff
  File "/venv/lib/python3.8/site-packages/numpy/lib/function_base.py", line 1447, in diff
    a = op(a[slice1], a[slice2])
TypeError: unsupported operand type(s) for -: 'tuple' and 'tuple'

```

Sameep2808:
@DanilYachmenev Please check this 
> @DanilYachmenev Just found out that `dassert_increasing_index()` cannot handle multiindex dataframe it fails at [this line](https://github.com/sorrentum/sorrentum/blob/52821f89bc16fecb4e0f67aca12f7e5c7baa8913/helpers/hpandas.py#L146) with the following error:
> 
> ```
> ______________________ Test_dassert_time_indexed_df.test4 ______________________
> Traceback (most recent call last):
>   File "/app/helpers/test/test_hpandas.py", line 3396, in test4
>     hpandas.dassert_time_indexed_df(df, allow_empty, strictly_increasing)
>   File "/app/helpers/hpandas.py", line 229, in dassert_time_indexed_df
>     dassert_increasing_index(df)
>   File "/app/helpers/hpandas.py", line 155, in dassert_increasing_index
>     mask = np.diff(index) <= pd.Timedelta(seconds=0)
>   File "<__array_function__ internals>", line 180, in diff
>   File "/venv/lib/python3.8/site-packages/numpy/lib/function_base.py", line 1447, in diff
>     a = op(a[slice1], a[slice2])
> TypeError: unsupported operand type(s) for -: 'tuple' and 'tuple'
> ```

helpers/test/test_hpandas.py
DanilYachmenev: implement

helpers/test/test_hpandas.py
DanilYachmenev: Hmm I see
Well, it means that `dassert_time_indexed_df()` is not really adjusted to work with multi-index dataframes
this means some changes to the function itself is required and it would be better to separate the changes to different PRs

therefore, lets remove all the changes related to `dassert_time_indexed_df()` from this PR

helpers/hpandas.py
DanilYachmenev, line 237: and well, it does not really robust fix compare to the changes in `dassert_index_is_datetime()`
@tamriq do we need this function to process multi-index dfs?
if yes, wdyt about making it more robust?
because otherwise we encounter https://github.com/sorrentum/sorrentum/pull/548#discussion_r1306022539
Can be a task for @Sameep2808 as well

but lets do it if it is needed only

helpers/test/test_hpandas.py
Sameep2808: I did, lint is overwriting this for some reason

helpers/hpandas.py
```diff
@@ -98,9 +98,18 @@ def dassert_index_is_datetime(
...
     index = _get_index(obj)
-    hdbg.dassert_isinstance(index, pd.DatetimeIndex, msg, *args)
+    if isinstance(index, pd.MultiIndex):
+        # In case of multi index check that at least one level is a datetime.
+        is_any_datetime = any(
+            isinstance(level, pd.DatetimeIndex) for level in index.levels
+        )
+        hdbg.dassert(is_any_datetime, msg, *args)
+    else:
+        hdbg.dassert_isinstance(index, pd.DatetimeIndex, msg, *args)
```
DanilYachmenev, line 112: comment for integrators:

this code is moved from cmamp master

helpers/test/test_hpandas.py
```diff
@@ -3232,3 +3233,81 @@ def test3(self) -> None:
...
+        """
+        Check that single-indexed dataframe index is datetime type.
+        """
+        index_is_datetime = True
+        df = self.helper(index_is_datetime)
+        df = df.loc["index1"]
+        hpandas.dassert_index_is_datetime(df)
+
+
+# #############################################################################
```
DanilYachmenev, line 3313: no need for the separator at the end of the file
just leave 1 empty line after the last line of code

DanilYachmenev (commented):
pls fix the conflict and 1 nit
almost there

DanilYachmenev (commented):
lgtm

helpers/hpandas.py
```diff
@@ -98,7 +98,7 @@ def dassert_index_is_datetime(
     """
     Ensure that the dataframe has an index containing datetimes.
 
-    It works for both single and multi-indexed dataframes.
+    Works for both single and multi-indexed dataframes.
```
gpsaggese, line 101: Let's leave the `it`. We use complete English phrases or imperative when PEP8 requires it.

helpers/test/test_hpandas.py
```diff
@@ -3272,3 +3273,78 @@ def test3(self) -> None:
...
         """
         self.assert_equal(act, exp, fuzzy_match=True)
+
+
+# #############################################################################
+
+
+class Test_dassert_index_is_datetime(hunitest.TestCase):
+    @staticmethod
+    def helper(
```
gpsaggese, line 3283: let's call it `get_test_df`

gpsaggese (commented):
Minor things and then merge

Sameep2808:
@DanilYachmenev Any updates on merging and closing the issue?

helpers/test/test_hpandas.py
```diff
@@ -3272,3 +3273,78 @@ def test3(self) -> None:
...
+# #############################################################################
+
+
+class Test_dassert_index_is_datetime(hunitest.TestCase):
+    @staticmethod
+    def get_test_df(
+        index_is_datetime: bool,
+    ) -> pd.DataFrame:
+        """
+        Get multi-index dataframe to unit test the assert.
```
gpsaggese, line 3287: Nit. You are introducing a dependency in the comment that is not needed.
If you move the function, then you need to update also the comment (which probably one will forget).

I would call the function `get_multiindex_df` and then put in the docstring an example of how the output looks like. You can grep in the code to see example of this.

gpsaggese (commented):
Nice code. 1 nit

gpsaggese (commented):
LG

helpers/test/test_hpandas.py
```diff
@@ -3272,3 +3273,92 @@ def test3(self) -> None:
...
+                2022-01-01 21:10:00+00:00   1.303778    -0.288235
+        index2  2022-01-01 21:00:00+00:00   1.237079    1.168012
+                2022-01-01 21:10:00+00:00   1.333692    1.708455
+        Example of dataframe returned when index_is_datetime = False.
+                            column1     column2
+        index   timestamp                    
+        index1  string1     -0.122140   -1.949431
+                string2     1.303778    -0.288235
+        index2  string1     1.237079    1.168012
+                string2     1.333692    1.708455
```
DanilYachmenev, line 3301: lets maintain empty line after the main docstring message and between the examples + put code pieces in backticks:
```
        """
        Helper function to get test multi-index dataframe.

        Example of dataframe returned when `index_is_datetime = True`:
                                            column1     column2
        index   timestamp                                    
        index1  2022-01-01 21:00:00+00:00   -0.122140   -1.949431
                2022-01-01 21:10:00+00:00   1.303778    -0.288235
        index2  2022-01-01 21:00:00+00:00   1.237079    1.168012
                2022-01-01 21:10:00+00:00   1.333692    1.708455

        Example of dataframe returned when `index_is_datetime = False`:
                            column1     column2
        index   timestamp                    
        index1  string1     -0.122140   -1.949431
                string2     1.303778    -0.288235
        index2  string1     1.237079    1.168012
                string2     1.333692    1.708455
        """
```

DanilYachmenev (commented):
1 more nit pls

DanilYachmenev:
@Sameep2808, could you pls move this changes to `cmamp`?
this is needed to keep repos integrated

Just make a copy of this PR there and put me as a reviewer

Sameep2808:
> @Sameep2808, could you pls move this changes to `cmamp`? this is needed to keep repos integrated

Sure will do add it today!

DanilYachmenev:
Integration PR
https://github.com/cryptokaizen/cmamp/pull/5574

Issue #538, opened 2023-08-14
Centralize input validation when running ETL scripts


DanilYachmenev:
We need to centralize checking logical dependencies between values of various arguments for ETL scripts.

To start we need to:
- [x] Create `im_v2/common/data/qa/validate_input_args.py` to store the function for the task
- [x] Create `validate_vendor_arg(vendor: str, args: argparse.Namespace) -> ivcdexex.Extractor` in the newly created file that will be doing what is done now from https://github.com/sorrentum/sorrentum/blob/master/im_v2/common/data/extract/download_bulk.py#L77 up to line 95
- [x] Replace the mentioned code by the newly created `validate_vendor_arg()`

FYI @jsmerix

samarth9008:
Please assign an EPIC to it.

gpsaggese:
@samarth9008

Two refinements of the rules:
1) let's ignore the EPICs in sorrentum, and use EPICs only in cmamp
2) When an issue is closed without an EPIC, we reopen it so that we don't forget to set EPIC and estimate.
Of course, let's only do it moving fwd and not retroactively.

Can you pls update the EpicMeister doc with these 2 rules?

PR #552, opened 2023-08-24
Sorr task538 centralize input validation when running etl scripts

Sameep2808:
Task #538 @DanilYachmenev I have created the new file and transferred the functions as per the issue


im_v2/common/data/qa/validate_input_args.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
```
DanilYachmenev, line 1: remove
this one is needed only for scripts, and this is a lib file

im_v2/common/data/qa/validate_input_args.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+Centralizing checking logical dependencies between values of various arguments
```
DanilYachmenev, line 3: -> `Centralize ...`

im_v2/common/data/qa/validate_input_args.py
```diff
@@ -0,0 +1,48 @@
...
+import im_v2.crypto_chassis.data.extract.extractor as imvccdexex
+
+
+def validate_vendor_arg(
+    vendor: str, args: argparse.Namespace
+) -> ivcdexex.Extractor:
+    """
+    Function to validate vendors arguments for contract type.
+
+    :param vendor: string to get the vendor
```
DanilYachmenev, line 25: -> `:param vendor: vendor name`

im_v2/common/data/qa/validate_input_args.py
```diff
@@ -0,0 +1,48 @@
...
+    :param args: args to get the contract type
+    """
```
DanilYachmenev, line 27: add also `:return: vendor data extractor`

im_v2/common/data/qa/validate_input_args.py
```diff
@@ -0,0 +1,48 @@
...
+import argparse
+
+import helpers.hdbg as hdbg
+import im_v2.binance.data.extract.extractor as imvbdexex
+import im_v2.ccxt.data.extract.extractor as imvcdexex
+import im_v2.common.data.extract.extractor as ivcdexex
+import im_v2.crypto_chassis.data.extract.extractor as imvccdexex
+
+
+def validate_vendor_arg(
```
DanilYachmenev, line 19: @jsmerix should we maybe give it a different name?
it seems that this function now is rather about getting extractor than just validation

DanilYachmenev (commented):
sg, couple moments

im_v2/common/data/qa/validate_input_args.py
jsmerix: If this module is for validation then only validations (assertions) belong here. 
This function should only check:
- if vendor belongs to one of the three supported
- If `universe_part` is included when  the vendor arg is `crypto_chassis`

Having a function which returns the correct instance is also useful but should belong to a different module, okay for me to file a follow-up.

jsmerix (changes requested):

im_v2/common/data/qa/validate_input_args.py
DanilYachmenev: git it, thx

```
def validate_vendor_arg(vendor: str, args: argparse.Namespace) -> None:
    """
    Validate vendor args.
    
    :param vendor: vendor name
    :param args: ETL script args
    """
    hdbg.dassert_in(vendor, ["binance", "ccxt", "crypto_chassis"], msg=f"Vendor {vendor} is not supported.")
    if vendor == "crypto_chassis":
        if not args.get("universe_part"):
            raise RuntimeError(
                f"--universe_part argument is mandatory for {vendor}"
            )
```

@Sameep2808 lets implement the changes here
update the script accordingly (run this function first and then init extractor)

Sameep2808:
@DanilYachmenev did the required changes you can have a look now

jsmerix (approved):

im_v2/common/data/extract/download_bulk.py
```diff
@@ -92,8 +90,6 @@ def _run(args: argparse.Namespace) -> None:
             # TODO(Vlad): Temporary stick to daily data for Binance.
             time_period=imvbdexex.BinanceNativeTimePeriod.DAILY,
         )
-    else:
-        hdbg.dfatal(f"Vendor {vendor} is not supported.")
```
DanilYachmenev, line 96: lets keep these lines
`if` clauses should always be completed with `else`
https://github.com/sorrentum/sorrentum/blob/master/docs/Coding_Style_Guide.md#make-if-elif-else-complete

DanilYachmenev (commented):
1 nit and lgtm

Sameep2808:
> lets keep these lines `if` clauses should always be completed with `else` https://github.com/sorrentum/sorrentum/blob/master/docs/Coding_Style_Guide.md#make-if-elif-else-complete

@DanilYachmenev I just thought that would have been a DRY with [this](https://github.com/sorrentum/sorrentum/pull/552/files#:~:text=else%3A,not%20supported.%22)

Sameep2808:
> lets keep these lines `if` clauses should always be completed with `else` https://github.com/sorrentum/sorrentum/blob/master/docs/Coding_Style_Guide.md#make-if-elif-else-complete

I thought that `else` would be DRY with [this](https://github.com/sorrentum/sorrentum/pull/552/files#:~:text=hdbg.dassert_in,not%20supported.%22%2C)

gpsaggese:
> > lets keep these lines `if` clauses should always be completed with `else` https://github.com/sorrentum/sorrentum/blob/master/docs/Coding_Style_Guide.md#make-if-elif-else-complete
> 
> I thought that `else` would be DRY with [this](https://github.com/sorrentum/sorrentum/pull/552/files#:~:text=hdbg.dassert_in,not%20supported.%22%2C)

This is a valid point, but in this case the DRY principle succumbs to keeping the code maintainable. 

I would write the code like:

```python
    hdbg.dassert_in(
        vendor,
        ["binance", "ccxt", "crypto_chassis"],
        msg=f"Vendor {vendor} is not supported.",
    )
    if vendor == "crypto_chassis":
        if not args.get("universe_part"):
            raise RuntimeError(
                f"--universe_part argument is mandatory for {vendor}"
            )
    elif vendor == "ccxt":
        # Do something or raise NotImplementedError
    elif vendor == "crypto_chassis":
        # Do something or raise NotImplementedError
    else:
        raise ValueError("Invalid error vendor='${vendor}'")
```

Assume that you add another vendor, then we centralize the assertion and add a new vendor to assertion.
Then we need to go and update all the if-then-else everywhere.
Surely we will miss one. With your solution you will get a silent error (return a `None`), the solution above will at least assert and you can find that there is a missing branch.

If the code in each branch I would actually remove the `dassert`.
The `dassert` is mainly to fail quickly, but it is redundant. So you need a way to keep the redundancy in sync, hence the `else`.

From docs/Reading_List.md
```
# Coding

- [The Pragmatic Programmer](https://www.amazon.com/Pragmatic-Programmer-Journeyman-Master/dp/020161622X)
  - Aka the Black Book
  - Reading (and really internalizing) this book is equivalent to the experience
    you can gain by coding for 20 years
  - It will change your life

- [Code Complete: A Practical Handbook of Software Construction, Second
  Edition](https://www.amazon.com/Code-Complete-Practical-Handbook-Construction/dp/0735619670)
  - Likely the only decent product ever made by Micro$oft
```

Sameep2808:
> Assume that you add another vendor, then we centralize the assertion and add a new vendor to assertion.
> Then we need to go and update all the if-then-else everywhere.
> Surely we will miss one. With your solution you will get a silent error (return a `None`), the solution above will at least assert and you can find that there is a missing branch.

> If the code in each branch I would actually remove the `dassert`.
> The `dassert` is mainly to fail quickly, but it is redundant. So you need a way to keep the redundancy in sync, hence the `else`.

@gpsaggese Thanks for clarifying I understood why the `else` should be implemented. Will keep this in mind for the future implementation.

Just one clarification. 

> ```python
>     elif vendor == "ccxt":
>         # Do something or raise NotImplementedError
>     elif vendor == "crypto_chassis":
>         # Do something or raise NotImplementedError
>     else:
>         raise ValueError("Invalid error vendor='${vendor}'")
> ```

The above part is already implemented [here](https://github.com/sorrentum/sorrentum/blob/edbd2a2133ace3e5e250a93b6d3f06826f80b64a/im_v2/common/data/extract/download_bulk.py#L82C9-L96C58) just after this function is called. As discussed earlier in this issue this function that we are implementing is just for validation purpose. And hence the remaining else if were transferred back to `im_v2/common/data/extract/download_bulk.p`. But do let me know if you want me to transfer it back and I'll make the needed changes.

DanilYachmenev (commented):
lgtm

gpsaggese:
Let's make the changes for this bug and I've filed a follow up bug #557

jsmerix (approved):

PR #556, opened 2023-08-28
fix_fast_tests

rheenina:


helpers/test/test_lib_tasks.py
```diff
@@ -133,7 +133,7 @@ def _cmd(cmd):
 
 # TODO(ShaopengZ): fails when running Sorrentum on ck server. `gh auth login`
 # issue.
-# @pytest.mark.skip
+@pytest.mark.skip
```
PomazkinG, line 136: let's consult with @LibertasSpZ before completely skipping.

does it pass:
a) when running Sorrentum from a personal laptop?
b) from CK dev server?

same for the `test_get_gh_issue_title4` below

LibertasSpZ: Thanks @PomazkinG .
a> If I recall correctly, on a personal laptop after user makes sure they have proper `gh auth status`, they pass. 
b> I just tried running them on `dev2/src/cmamp1`. After asking for username and password twice, they pass.
c> They failed the`gh` workflow, at least when run on Sorrentum, for `gh auth login` issues. [Presumably because they are not mocked and the workflow cannot fill in the login information].

PomazkinG: @LibertasSpZ thanks! To re-iterate: it works everywhere but not when is run via GH actions in sorrentum? do we have a bug for the `gh auth login` sorrentum issue?

LibertasSpZ: @PomazkinG  Correct. 
I don't see a bug about `gh auth login` in Issues. Please feel free to file a bug. Thanks!

PomazkinG: @LibertasSpZ sg, filed https://github.com/cryptokaizen/cmamp/issues/5211, feel free to add details

gpsaggese: Yes I think there are corner cases as the ones brought up.

The matrix of builds (x86 vs arm, Sorrentum vs cmamp, laptop vs dev vs CI) is too complicated.
So we support only x86 x cmamp x {dev, CI}.
For all the other builds it's ok to put a TODO with some details of the problem and a skip.

@PomazkinG feel free to update the build documentation with these heuristics.

LibertasSpZ: Sorry, correction: on `dev2:src/cmamp`, the tests that ask for username and password twice were`test_git_branch_create1()` and `test_git_branch_create2()`, not `test_get_gh_issue_title4` and `test_gh_login()`.

PR #558, opened 2023-08-28
SorrTask_Make_changes_from_Integration

samarth9008:
- Changes as per
   - https://github.com/sorrentum/sorrentum/pull/535#discussion_r1307979151
   - https://github.com/cryptokaizen/cmamp/pull/5012#discussion_r1307941473

### DIFF
--- a/im_v2/ccxt/data/client/test/test_ccxt_clients.py
+++ b/im_v2/ccxt/data/client/test/test_ccxt_clients.py
@@ -1284,10 +1284,6 @@ class TestCcxtHistoricalPqByTileClient1(icdc.ImClientTestCase):
 
     # TODO(gp): Difference between cmamp1 and sorrentum. Why is there a
     # difference?
-    @pytest.mark.skipif(
-        not henv.execute_repo_config_code("get_name()") == "//amp",
-        reason="Run only in //amp",
-    )
     @pytest.mark.slow("Slow via GH, but fast on the server")
     def test_filter_columns1(self) -> None:
         resample_1min = True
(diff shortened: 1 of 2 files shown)


im_v2/ccxt/data/client/test/test_ccxt_clients.py
```diff
@@ -1284,10 +1284,6 @@ def test_read_data7(self) -> None:
 
     # TODO(gp): Difference between cmamp1 and sorrentum. Why is there a
     # difference?
-    @pytest.mark.skipif(
```
gpsaggese, line 1287: Can we find a fix that has the same exact code but works differently for the two repos?
E.g., `get_name()` returns `//sorr` vs `//amp` so we should be able to express the condition with the same code, executing differently in two repos.

Then we can remove the TODO, or actually we should say something like
```
TODO(gp): Figure out why this test doesn't pass in ...
```

Makes sense?

samarth9008:
@rheenina The 38ac6a1720bab8a5e05909f31a54ebd40fc754f5 commit has the implementation of https://github.com/cryptokaizen/cmamp/issues/5211#issuecomment-1696516915 and the tests are skipped on `sorrentum` repo.

im_v2/ccxt/data/client/test/test_ccxt_clients.py
samarth9008: Not exactly. What is that we are trying to solve? Also from the `TODO`, the tests are passing in both the repos.

PomazkinG: I guess we can remove the TODO `    # TODO(gp): Difference between cmamp1 and sorrentum. Why is there a
    # difference?` and all good here

PomazkinG: This PR resolves the TODO

PomazkinG (commented):
LG here

do we need to move the pytest timeout fix here? from https://github.com/cryptokaizen/cmamp/pull/5212/files

PomazkinG:
Good work @samarth9008, merging

PR #569, opened 2023-09-15
CmTask5422_Fixing_Pytest_marker

samarth9008:
https://github.com/cryptokaizen/cmamp/issues/5422

- Fxing pytest marker to fix the test.


samarth9008:
Merging since I just removed the `msg` from pytest skip marker and it will block all other merges too. If any changes, then I will file a follow up PR.

Issue #539, opened 2023-08-14
LinkedIn Flow follow up


yiyunlei:
PR #525 follow up.
- Describe the updated flow in the gdoc and request @gpsaggese review before implementing it

yiyunlei:
ETA: Friday(2023-08-18)
- Describe the updated flow in the gdoc.

gpsaggese:
@yiyunlei to finish this guy

yiyunlei:
These packages should be installed in the docker, otherwise, we need to install them manually and wait 1~2 minutes before using Google API. :)
```
google-auth google-auth-httplib2 google-auth-oauthlib google-api-python-client
gspread-pandas
```

gpsaggese:
Let's install manually for now in the notebook, since the container is already enormous
```
!source /venv/bin/activate; pip install ...
```

gpsaggese:
To export a port in Docker you need to
```
> git diff helpers/lib_tasks_docker.py
diff --git a/helpers/lib_tasks_docker.py b/helpers/lib_tasks_docker.py
index cb75b47ce..0cc21b929 100644
--- a/helpers/lib_tasks_docker.py
+++ b/helpers/lib_tasks_docker.py
@@ -640,6 +640,7 @@ def _generate_docker_compose_file(
           ports:
             # TODO(gp): Rename `AM_PORT`.
             - "${PORT}:${PORT}"
+            - 8181:8181

         # TODO(gp): For some reason the following doesn't work.
         #  jupyter_server_test:
```

I've also made some changes to the doc
[Gsheet_into_pandas.md](https://github.com/sorrentum/sorrentum/files/12564264/Gsheet_into_pandas.md)

Can you pls incorporate these changes in your PR?

yiyunlei:
OK, no problem.

why2000:
This should be closed as task specified here finished.

Issue #522, opened 2023-08-07
Follow up for PB API


yiyunlei:
Follow up for #515. 
Resolve the conversations in #515 PR later.
- Update to not use `hio.to_file()`

gpsaggese:
Very good tracking this.

You can also reuse the same #515 and just do a PR with a name like
`SorrTask487_Automate_extracting_the_PB_results_and_download_to_local_storage_2`

yiyunlei:
@gpsaggese @DanilYachmenev @samarth9008

Could you help me take a look at this issue?

https://github.com/sorrentum/sorrentum/pull/515#discussion_r1284895575

hio.to_file() seems like having an encoding issue.
After I used hio.to_file to save result CSV file, I got content like this:
```
â�¢        Dynamic global entrepreneur, visionary founder Global Coaches Network, a pioneer in the field of global leadership coaching with 300 coaches, operating in 75 countries
â�¢        25 years developing women organizational leaders and entrepreneurs to maximize their leadership potential   
â�¢        25 yearsâ�� experience developing global executives and their families in the US, Asia, and the Middle East as their coach, family business consultant, or their therapist. 
â�¢        Speaker and trainer on global leadership in the 21st century, global leadership coaching for coaches and leaders in Asia, Europe, the US, and Africa. 
â�¢        Award winning author of a series of articles on â��Ensuring Success on the International Assignmentâ��. 
â�¢        20 years marriage and family therapist
â�¢        Fun fact: Barrie was once a nanny for the Jordanian Royal family
```
Which should be:
```
•        Dynamic global entrepreneur, visionary founder Global Coaches Network, a pioneer in the field of global leadership coaching with 300 coaches, operating in 75 countries
•        25 years developing women organizational leaders and entrepreneurs to maximize their leadership potential   
•        25 years’ experience developing global executives and their families in the US, Asia, and the Middle East as their coach, family business consultant, or their therapist. 
•        Speaker and trainer on global leadership in the 21st century, global leadership coaching for coaches and leaders in Asia, Europe, the US, and Africa. 
•        Award winning author of a series of articles on “Ensuring Success on the International Assignment”. 
•        20 years marriage and family therapist
•        Fun fact: Barrie was once a nanny for the Jordanian Royal family
```

gpsaggese:
It’s ok not to use to_file in this case

yiyunlei:
@gpsaggese Thanks, will update it.

gpsaggese:
[...]

PR #566, opened 2023-09-12
 SorrTask416 Figure out permissions and gspread

yiyunlei:
Instructions:
https://docs.google.com/document/d/1dfAVMJZtHtQ3sX7UHwpXNAWPiJs17w0V8FlsOI7a32M/edit?usp=sharing

PR #574, opened 2023-09-25
Sorr task522 follow up for pb api

why2000:
#522 and #539 Moved google file apis, reframed it into helper methods instead of a class. Solved some minor bugs.


why2000:
@gpsaggese Since this PR requires a change to the docker compose builder, it will fail the check (it includes the changes in #566). So we may want to update the docker compose test if this one is to be merged.

why2000:
Refined the get credentials method to cover more situations. Updated the doc following [#539](https://github.com/sorrentum/sorrentum/issues/539#issuecomment-1712357259) and based on the new helper logic.

gpsaggese:
@PomazkinG can you PTAL to this PR?

Besides some nits, we need to:
1) Implement the mechanism to install gspread dynamically (like have a `def install_gspread_if_needed()`).
2) We want to open a port for letting Google authenticate, but we need to do this in the YAML generator of the invoke
- This is something @why2000 can address when he is in cmamp

gpsaggese (commented):
Edited a bit, but LG.

gpsaggese:
Can you move this to cmamp? We'll keep working on this there if needed

why2000:
> Can you move this to cmamp? We'll keep working on this there if needed

Will do later today.

why2000:
Now moved, see the cmamp PR above

Issue #612, opened 2023-11-27
Marketing: Get information from Signal


why2000:
https://docs.google.com/document/d/1z-svt4K6dSy5ncuVZzrtL9ZYuBmknCp8UfGfxljaBNQ/edit#heading=h.g37b603qcnlr

Get information from here.
https://signal.nfx.com/investor-lists/top-fintech-seed-investors

samarth9008:
Make sure you assigned epic and sprint

why2000:
When accessing the website too many times, we will not be able to see any inner info of the investors. It will require us to login before being able to view details. But with certain actions we can continue loading the investors list page and read more investors' name and company. I'll implment the script in this way.

gpsaggese:
@why2000 see convo on TG

PR #609, opened 2023-11-21
Sorr task601 followup fix

why2000:
#601 
FYI @samarth9008 @gpsaggese 
This is a follow up PR for #603 . 
* Simplified the workflow. Now the code can be applied directly on `.mht` or `.mhtml` files. No need to manual convert to html.
* Restructured the code into two libraries, and the notebooks using them. Both moved to `marketing` folder to be consistent with #606 .
* One thing to be marked up (also put in the comment): We won't be able to get any info if we download the webpage directly into an `html`. An embedded safekeeper code will erase everything before it is downloaded as an html. Which means `.mht` file format is a must-have when downloading it. Previously due to BeautifulSoup compatibility, we'll need to manually convert `mht` to `html` through a browser, now by applying the `quoted-printable` format decoder, it is no longer needed.

PR #614, opened 2023-11-28
Sorr task612 get information from signal

why2000:
https://docs.google.com/document/d/1z-svt4K6dSy5ncuVZzrtL9ZYuBmknCp8UfGfxljaBNQ/edit#heading=h.g37b603qcnlr
#612 
FYI @samarth9008 @gpsaggese 
This PR is using the structure from #609 and based on it. Better to review after merging that PR.


marketing/signal/extract_investors_from_signal_list.py
```diff
@@ -0,0 +1,95 @@
+"""
+Import as:
+
+import marketing.signal.extract_investors_from_signal_list as mseifsili
+"""
+
+# !sudo /bin/bash -c "(source /venv/bin/activate; pip install --upgrade selenium webdriver-manager)"
```
samarth9008, line 7: Remove it if not used.

marketing/notebooks/SorrTask612_Get_information_from_Signal.py
```diff
@@ -0,0 +1,34 @@
...
+# %%
+import marketing.signal.extract_investors_from_signal_list as mseifsili
+
+# %% run_control={"marked": true}
+# Destination result file path.
+investors_csv_save_path = "../result_csv/signal_fintech_investors.csv"
+# Source data page URL.
+baseurl = "https://signal.nfx.com/investor-lists/top-fintech-seed-investors"
+start_idx = 0
+length = 40
```
samarth9008, line 28: Can you explain why are these required?

marketing/signal/extract_investors_from_signal_list.py
```diff
@@ -0,0 +1,95 @@
...
+    baseurl: str, start_idx: int, length: int
+) -> pd.DataFrame:
+    """
+    Extract a dataframe of investor information from a signal investors list
+    page. e.g. https://signal.nfx.com/investor-lists/top-fintech-seed-investors
+    Available lists are in this page: https://signal.nfx.com/investor-lists/
+
+    :param baseurl: str, the page url to be extracted
+    :param start_idx: int, the index of the first item to be extracted (start from 0)
+    :param length: int, the number of items to be extracted
```
samarth9008, line 30: No types in param description.

samarth9008 (commented):
I see there are errors in your notebook. Is there any reason for it?

marketing/signal/extract_investors_from_signal_list.py
why2000: The error in notebook is because I'm using WSL and selenium cannot open the browser under WSL jupyter notebook. I've tested this on an macbook and it works fine. For this line: We'll need to install those dependencies to get the code running, since selenium is not included in the image.

marketing/notebooks/SorrTask612_Get_information_from_Signal.py
why2000: This is because the page is not loading all items at first. I'm not sure of the total length of it, so I want to give a range of data to load, thus it won't have a too long waiting time.

gpsaggese (commented):
Henry is officially the scraping ninja on the team.
I can't add much, besides let's make sure to document things both in code and outside.

I'll let @samarth9008 work with Henry on this stuff.

marketing/signal/extract_investors_from_signal_list.py
samarth9008: A better thing would be to add a comment above it regarding its usage since this is not used anywhere in this file but will be used in another file which will make use of this library.

marketing/notebooks/SorrTask612_Get_information_from_Signal.py
samarth9008: Please mention this in the comment.

marketing/notebooks/SorrTask601_Extract_VCs_from_Tra_search_mhtml.py
```diff
@@ -0,0 +1,40 @@
...
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
+# %% run_control={"marked": true}
+import marketing.tra.extract_VCs_from_Tra_search_mhtml as mtevftsmh
+
+# %% [markdown]
+# # Workflow before using this library:
```
gpsaggese, line 19: Good explanation.

Can we move to our new approach to doc following diataxis.fr ?
 
Like an `marketing.explanation.md` and `tra.how_to_guide.md`?
@samarth9008 can you help? We can do a PR with the outline of the doc and then fill it up

marketing/signal/extract_investors_from_signal_list.py
```diff
@@ -0,0 +1,102 @@
+"""
+Import as:
+
+import marketing.signal.extract_investors_from_signal_list as mseifsili
+"""
+
+# TODO(Henry): This package need to be manually installed until they are added
```
gpsaggese, line 7: it is added

just to show that I do read the code :-)

gpsaggese (commented):
Good work. Let's merge and then we can do a follow up.

marketing/notebooks/SorrTask601_Extract_VCs_from_Tra_search_mhtml.py
samarth9008: @why2000 

Instead of writing this in a notebook, we would like to create a md doc like `tra.how_to_guide.md`, explaining its workflow end-to-end. Toe location will be `marketting/tra`

Same for signal and for @yiyunlei same for `dropcontact` if applicable. Location will be there respective folders.

marketing/signal/extract_investors_from_signal_list.py
why2000: Which version of images are they added? I still cannot import selenium on dev server's notebook yet. Also `webdriver-manager` needs to be added together for selenium to work: we don't want to download webdriver manually on server. It won't bundle any driver binary inside, but to auto-determine which version of driver to be downloaded when we start selenium.

PR #608, opened 2023-11-18
SorrTask606 Extract emails using DropContact

yiyunlei:
- PR for #606 

- Test result:
https://docs.google.com/spreadsheets/d/1MIw-R17PikYJm2_0eNpMna8STNoDJYL6kAfZs8qY0m0/edit#gid=1404487521

PR #616, opened 2023-11-28
SorrTask606 Follow up Get email from dropcontact

yiyunlei:
Follow up:
- #608


marketing/dropcontact/find_email.py
```diff
@@ -256,58 +201,14 @@ def get_email_from_dropcontact(
     :param last_name_list: List of last names.
     :param company_list: List of company names.
     :param api_key: API key of DropContact.
-    :return: A dataframe with columns: first name, last name, full name, email, phone, pronoun, job title.
+    :return: A dataframe with the following columns:
+            first name, last name, full name, email, phone, pronoun, job title.
```
samarth9008, line 205: Not dots at the end. Remove it from all the functions.

samarth9008 (approved):
LGTM

marketing/dropcontact/find_email.py
gpsaggese: Yes that's the rule. Probably we'll add a rule to the linter since it's difficult to keep everyone in sync here.

marketing/dropcontact/find_email.py
```diff
@@ -1,92 +1,37 @@
...
-# # DropContact functions
-
-# %%
-def preprocess_data(
-    first_name_list: Iterable[str],
-    last_name_list: Iterable[str],
-    company_list: Iterable[str],
-) -> List:
+def _preprocess_dropcontact_data(
+    first_name_list: Sequence[str],
```
gpsaggese, line 18: I agree that this is the right way, but personally I tend to just stick to List[str] if it's a list.

I would call `first_name_list` -> `first_names` to avoid the Hungarian notation.

gpsaggese (commented):
LG.

On one side, I'm ok with being a bit more relaxed with following the rules since this is research code. On the other side, learning all our rules is useful to follow our style.

We want to automate more and more of the formatting so that we don't spend time thinking about those.

I'll let @samarth9008 decide the right balance btw making progress and respect the rules.

marketing/dropcontact/find_email.py
yiyunlei: Name changed. But this is not necessarily a List, it can be anything that has a length, including a pandas column sequence.

Issue #631, opened 2024-01-18
Improve chatGPT assistant integration


why2000:
This is a follow-up of #624
Fixing the conversations and adding docstrings for `helpers/hchatgpt.py`
Also, as mentioned here https://github.com/sorrentum/sorrentum/pull/625#discussion_r1446024333 we may want to have `openai` installed in our image. 
A detailed usage doc is created at `docs/work_tools/all.chatgpt_api.how_to_guide.md`

gpsaggese:
No need to complicate the container flow, let's just leave the try-catch-install since it works fine

gpsaggese:
Next steps:
- Move to cmamp
- Comments from the PR

why2000:
FYI @samarth9008 
This issue is already done, and it is integrated into cmamp in AmpTask1786

PR #634, opened 2024-01-19
Bugfix for empty model

why2000:
#631 
@gpsaggese a quick fix as mentioned in tg.
Fixed a bug where code will crash if no model is given. Now default model is correctly set as `gpt-3.5-turbo-1106`


gpsaggese:
Pls mirror the change to cmamp and then let's switch to developing this in cmamp.

PR #638, opened 2024-01-22
CmTask_update_submoduel

samarth9008:
https://github.com/sorrentum/dev_tools/issues/539#issuecomment-1900529831

PR in dev_tools - https://github.com/sorrentum/dev_tools/pull/546

### DIFF
--- a/helpers/hgit.py
+++ b/helpers/hgit.py
@@ -541,7 +541,7 @@ def _get_repo_short_to_full_name(include_host_name: bool) -> Dict[str, str]:
     # From short name to long name.
     repo_map = {
         "amp": "alphamatic/amp",
-        "dev_tools": "alphamatic/dev_tools",
+        "dev_tools": "sorrentum/dev_tools",
     }
     if include_host_name:
         host_name = "github.com"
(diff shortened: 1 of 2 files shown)


helpers/hgit.py
```diff
@@ -541,7 +541,7 @@ def _get_repo_short_to_full_name(include_host_name: bool) -> Dict[str, str]:
     # From short name to long name.
     repo_map = {
         "amp": "alphamatic/amp",
-        "dev_tools": "alphamatic/dev_tools",
+        "dev_tools": "sorrentum/dev_tools",
```
gpsaggese, line 544: Sg, although this might be a pain in the neck for the integration, since one needs to be careful in leaving this diff btw repos.

Ideally we would want to get this as function of `repo_config`. I think there is a mechanism to do it with something like `henv.execute_repo_config_code("get_html_bucket_path()")`

helpers/test/test_git.py
```diff
@@ -167,7 +166,7 @@ def test_get_repo_name2(self) -> None:
         self.assert_equal(act, exp)
 
     def test_get_repo_name4(self) -> None:
-        full_name = "alphamatic/dev_tools"
+        full_name = "sorrentum/dev_tools"
```
gpsaggese, line 169: Same here.

gpsaggese (commented):
Good work. One suggestion

helpers/hgit.py
samarth9008: We can do that by adding this path to `get_repo_map()` in repo_config.py but I would suggest to keep this in sync with the cmamp for now. wdyt? Also I do not have a clear mental model yet w.r.t repo config file.

PR #649, opened 2024-02-06
Conventions file location changed

Shaunak01:
The location is changed and the file is renamed to docs/work_tools/all.basic_sorrentum_development.how_to_guide.md

@samarth9008


samarth9008 (changes requested):
The file from the previous location should also be deleted and pushed to this PR. We do not want same file with different names on different locations.

samarth9008 (approved):

PR #653, opened 2024-02-15
AmpTask1786_Integrate_20240215

gpsaggese:


samarth9008:
Will need to re-work properly on this PR as many tmp files are pushed which should not be the case. 

I believe would be more efficient to do a complete integration in a single PR if possible.

helpers/hunit_test.py
```diff
@@ -161,39 +161,10 @@ def convert_df_to_string(
...
-        decimals,
-    ):
-        n_rows = n_rows or len(df)
-        # Add N top rows.
-        output.append(df.head(n_rows).to_string(index=index))
-    # Convert into string.
-    output_str = "\n".join(output)
-    return output_str
+    handle_signed_zeros = True
+    return hpandas.df_to_str(df, handle_signed_zeros=handle_signed_zeros)
```
samarth9008, line 167: This whole function can be removed. As not used anymore.

gpsaggese: Agree. I'll do a clean up in the next integration, now that we are close to parity.

PR #671, opened 2024-02-22
CmTask7274: Update helpers

sonniki:
Updating helpers for the sake of https://github.com/sorrentum/dev_tools/pull/558


helpers/hstring.py
```diff
@@ -78,7 +78,7 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
     :return: the indices of docstrings
     """
     docstring_line_indices = []
-    quotes = {'"""': False, "'''": False}
+    quotes = {'"""': False, "'''": False, "```": False}
```
samarth9008, line 81: Why this change?

samarth9008 (commented):
Unrelated but can you remove this whole function as well. It is not being used anymore

```
helpers/hunit_test.py:157:def convert_df_to_string(
```

helpers/hgit.py
```diff
@@ -235,7 +235,7 @@ def is_in_amp_as_supermodule() -> bool:
     Return whether we are in the `amp` repo and it's a super-module, i.e.,
     `amp` by itself.
     """
-    return is_amp() and not is_inside_submodule(".")
+    return (_is_repo("amp") or _is_repo("cmamp")) and not is_inside_submodule(".")
```
gpsaggese, line 238: IMO we should leave `is_amp` here.

What would be the problem?

sonniki: It caused some tests to fail, see https://github.com/sorrentum/sorrentum/actions/runs/8008692610/job/21875671063.
Those failing tests are all skipped if `is_in_amp_as_supermodule()` returns False. But with the addition of `_is_repo("sorr")` in `is_amp()` it started returning True and the tests ultimately failed in this PR.

gpsaggese: My take is that the test goldens are incorrect

image

Let me look into that directly

samarth9008:
TODO(Samarth): Make a PR in sorrentum with changes in hgit file

samarth9008:
@gpsaggese Need to make a lot of other changes as the base PR is now sorrentum and not cmamp

gpsaggese (commented):
LG

PR #673, opened 2024-02-23
CmampTask7280_aws_CLI_doesn_t_work_in_Sorrentum_dev_tools_ARM_container

dremdem:
#7280

- mark some tests as `requires_ck_infra` since they need access to the AWS S3 resources


PomazkinG (commented):
General observation: since `ck_infra` is a broad term which may include S3, DB, etc., it would be time to explain which part of `ck_infra` a test needs

e.g.,
```
@pytest.mark.requires_ck_infra("Requires access to CK unit-test S3 bucket")
```

also if the marker is applicable to the whole class, it's better to mark the whole class instead of marking each test separately

gpsaggese (commented):
Looks good. 

FYI you can use the 
@pytest.mark.requires_ck_infra also for a class if simpler, instead of by test method.

Your call

samarth9008 (commented):
LG for now.

Did you run the tests without adding any aws credentials in your local mac? If so, these are the only tests that requires aws infra to pass??

If not, lets remove the aws credentials and run the fast tests on sorrentum. See which are failing and we can add markers accordingly to those as well

The goal is to mimic a scenario when someone who doesn't have access to dev server or any kind of aws infra should have no failures when running the tests locally.


We do not want these kind of scenario - https://github.com/sorrentum/sorrentum/issues/678#issuecomment-1963197730

- [x] Check aws arch on new image when running on local MAC
- [x] Run all the tests without any aws credentials configured
- [ ] Add appropriate markers to those tests.

LMK if you need a PP.

dremdem:
> * Add appropriate markers to those tests.

No new failing tests appear, so there's no need to add new markers.

gpsaggese:
1) Looks good to me. We could create more descriptive markers like `@pytest.mark.requires_ck_s3_bucket`, ... but P1 IMO.
2) Can we figure out why the GH Action is passing? We should not pass AWS credentials to mimic what people see on their setup. Makes sense?

samarth9008:
> Can we figure out why the GH Action is passing? We should not pass AWS credentials to mimic what people see on their setup. Makes sense?

@dremdem Do you have an idea related to it? Probably bcz of Github token??

dremdem:
> > Can we figure out why the GH Action is passing? We should not pass AWS credentials to mimic what people see on their setup. Makes sense?
> 
> @dremdem Do you have an idea related to it? Probably bcz of Github token??

I guess because of this part in the GH action workflow: 
image

We pass the creds from the AWS secrets to the environment variables.

PR #677, opened 2024-02-23
DevToolsTask550: Fix typo

sonniki:
Fixing a typo that leads to failed assertions.

### DIFF
--- a/dev_scripts/notebooks/add_toc_to_notebook.py
+++ b/dev_scripts/notebooks/add_toc_to_notebook.py
@@ -129,8 +129,8 @@ def _main(parser: argparse.ArgumentParser) -> None:
     if args.input_files:
         files_to_process = []
         for input_file in args.input_files:
-            hdbg.dassert_path_exists(args.input_file)
-            hdbg.dassert_file_extension(args.input_file, ".ipynb")
+            hdbg.dassert_path_exists(input_file)
+            hdbg.dassert_file_extension(input_file, ".ipynb")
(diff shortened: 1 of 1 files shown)


sonniki:
> Make sure to do the same PR in cmamp as well.

I thought we could do the same thing as with https://github.com/sorrentum/sorrentum/pull/671: I'll push amp in https://github.com/sorrentum/dev_tools/pull/562 and it will be integrated into cmamp at the next integration. Or will this not work?

samarth9008:
We changed the approach, if applicable any PR done on sorrentum should be mirrored on cmamp. The opposite will be done during integration. So that it will be only one-side integration from cmamp->sorrentum

sonniki:
> We changed the approach, if applicable any PR done on sorrentum should be mirrored on cmamp. The opposite will be done during integration. So that it will be only one-side integration from cmamp->sorrentum

Opened an identical PR in cmamp: https://github.com/cryptokaizen/cmamp/pull/7360
Do you think we can merge the cmamp PR and this one, before https://github.com/sorrentum/dev_tools/pull/562 is reviewed?
Just because reviewing https://github.com/sorrentum/dev_tools/pull/562 might take time while these typo/bug fixing PRs are really straightforward.

samarth9008:
We can merge in `cmamp` before not an issue. After merging in this, make sure to update amp in dev_tools PR so that the amp stays updated in that PR. Not merging rn before GP review's it

Issue #682, opened 2024-02-24
Improve hpandas.compare_dfs()


samarth9008:
Orginally from - https://github.com/cryptokaizen/cmamp/issues/5941

There are a couple of issues with `hpandas.compare_dfs()`

1) The function issues an assertion only when computing pct difference 
2) Assertion works only if we remove infinities
3) I feel like we are reinventing the wheel when trying to compare dfs, i.e. most of the functionality implemented already exists in [`numpy.isclose`](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html#numpy.isclose), not to mention that Numpy takes into account more edge cases

Some comparison invariants (copy-paste from https://numpy.org/doc/stable/reference/generated/numpy.isclose.html#numpy.isclose)

**Finite values:**

isclose uses the following equation to test whether two floating point values are equivalent.

`absolute(a - b) <= (atol + rtol * abs(b))`

**NaNs:**
NaNs are treated as equal if they are in the same place and if equal_nan=True.

**Infs:**

Infs are treated as equal if they are in the same place and of the same sign in both arrays.

So instead of

https://github.com/sorrentum/sorrentum/blob/476889c607fe17d72e574f56123366d1f2dc7df9/helpers/hpandas.py#L1990-L2000

and 

https://github.com/sorrentum/sorrentum/blob/476889c607fe17d72e574f56123366d1f2dc7df9/helpers/hpandas.py#L2015-L2024

we can use `numpy.allclose()` to check the equality before computing the diffs. E.g., seems like [`pd.testing.assert_frame_equal`](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html) is a wrapper around `np.isclose()` and on top of that it returns a nice assertion, e.g.,

```
DataFrame.iloc[:, 1] (column name="B") values are different (33.33333 %)
[index]: [0, 1, 2]
[left]:  [4, 5, 6]
[right]: [4, 5, 7]
At positional index 2, first diff: 6 != 7
```

The plan is:
1) Check if `assert_frame_equal` complies with the comparison invariants
2) We should add the `only_warning = True` switch to issue a warning on demand only
3) Try to use `assert_frame_equal()` inside the `compare_dfs()` and see if the regressions still work
4) Add more unit tests to ensure the nice coverage (ideally 100%)

Side notes:
[...]

PR #717, opened 2024-03-06
SorrTask682_improve_hpandas_compare_dfs

tkpratardan:
Task #682 
@samarth9008 @sonaalKant 

I have made changes and implemented the `pd.testing.assert_frame_equal()` in the required place. Please let me know if I am going in the right direction. Could you please review and let me know the required changes? 

I made a couple of changes to the logic of the code as well, which I felt needed to be fixed. Please let me know 

cc: @gpsaggese  

FYR:
The workflow for https://github.com/sorrentum/sorrentum/issues/682 is organized as follows:

1. Main PR: https://github.com/sorrentum/sorrentum/pull/717 (changes to the main function)
2. Stacked PR1: https://github.com/sorrentum/sorrentum/pull/733 (Scaffolding tests for hpandas.compare_dfs())
3. Stacked PR2: https://github.com/sorrentum/sorrentum/pull/726 (fixing linter changes in the files being affected, viz., helpers/hpandas.py and helpers/test/test_hpandas.py )


helpers/hpandas.py
```diff
@@ -761,8 +762,8 @@ def drop_axis_with_all_nans(
     """
     Remove columns and rows not containing information (e.g., with only nans).
 
-    The operation is not performed in place and the resulting df is returned.
```
tkpratardan, line 764: These have been changed after linter

helpers/hpandas.py
```diff
@@ -1934,6 +1937,7 @@ def compare_dfs(
     zero_vs_zero_is_zero: bool = True,
```
tkpratardan, line 1937: Do we require `compare_nans` variable? As `assert_frame_equal()` does not allow us to specify the use case

helpers/hpandas.py
```diff
@@ -1983,47 +1992,46 @@ def compare_dfs(
...
+            pd.testing.assert_frame_equal(df1, df2, check_like=True)
+        except AssertionError as e:
+            hdbg._dfatal(txt=e, msg=None, only_warning=only_warning)
+        finally:
+            df_diff = df1 - df2
+            if remove_inf:
+                df_diff = df_diff.replace([np.inf, -np.inf], np.nan)
     elif diff_mode == "pct_change":
-        df_diff = 100 * (df1 - df2) / df2
+        df_diff = 100 * (df1 - df2) / df2.abs()
```
tkpratardan, line 2010: added `df2.abs()`. Because in the previous version of the code, `df2 has -2` and `df1 has -1` the percentage change would be `-50` where it should have been `50`. Let me know if this right?

helpers/hpandas.py
```diff
@@ -1983,47 +1992,46 @@ def compare_dfs(
...
             df_diff[zero_vs_zero_mask] = 0
-    else:
-        raise ValueError(f"diff_mode={diff_mode}")
-    df_diff = df_diff.add_suffix(f".{diff_mode}")
-    if diff_mode == "pct_change" and assert_diff_threshold is not None:
-        # TODO(Grisha): generalize for the other modes.
-        # Report max diff.
-        max_diff = df_diff.abs().max().max()
-        _LOG.log(log_level, "Max difference factor: %s", max_diff)
+        if remove_inf:
```
tkpratardan, line 2017: Placed `remove_inf` before the checking for `assert_diff_threshold`  following the comments in the previous version. In the previous version the check would fail because `inf` is not removed, as it should be.

helpers/test/test_hpandas.py
```diff
@@ -2483,7 +2486,140 @@ def test8(self) -> None:
...
+            hpandas.compare_dfs(
+                df1, df2, compare_nans=compare_nans, only_warning=False
+            )
+
+    def test9(self) -> None:
+        """
+        Test to verify the error when df1 and df2 have different index types.
+        """
+        df1 = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})
+
```
gpsaggese, line 2498: No need for empty lines.

When you need an empty line is often an indication that you need to add a comment.

The unit tests should have a clear pattern

```
# Prepare inputs.
...
# Execute.
...
# Check output.
...
```

Can you file a bug to update ./docs/coding/all.write_unit_tests.how_to_guide.md with this
?

helpers/test/test_hpandas.py
```diff
@@ -2483,7 +2486,140 @@ def test8(self) -> None:
...
+        # Create df2 with a DatetimeIndex.
+        dates = pd.date_range("2021-01-01", periods=3)
+        df2 = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6], "timestamp": dates})
+        df2 = df2.set_index("timestamp")
+        with self.assertRaises(AssertionError):
```
gpsaggese, line 2503: Let's check the output of the assertion 

Something like:
helpers/test/test_dbg.py:670

Can you pls file an issue to you to adding this to 
./docs/coding/all.write_unit_tests.how_to_guide.md

You can mark this resolved posting the pointer to the GH Issue

helpers/test/test_hpandas.py
```diff
@@ -2483,7 +2486,140 @@ def test8(self) -> None:
...
+        timestamp
+        2022-01-01 21:01:00+00:00         0.001           0.001            0.001
+        2022-01-01 21:02:00+00:00         0.001           0.001            NaN
+        2022-01-01 21:03:00+00:00         0.001           0.001            0.001
+        """
+        self.assert_equal(actual, expected, fuzzy_match=True)
+
+    def test12(self) -> None:
+        """
+        -Check test case when negative values in df2.
```
gpsaggese, line 2589: No -

helpers/test/test_hpandas.py
```diff
@@ -3654,28 +3790,25 @@ def test5(self) -> None:
...
-        Helper function to get test multi-index dataframe.
-        Example of dataframe returned when `index_is_datetime = True`.
-                                            column1     column2
-        index   timestamp
-        index1  2022-01-01 21:00:00+00:00   -0.122140   -1.949431
-                2022-01-01 21:10:00+00:00   1.303778    -0.288235
-        index2  2022-01-01 21:00:00+00:00   1.237079    1.168012
-                2022-01-01 21:10:00+00:00   1.333692    1.708455
+        Helper function to get test multi-index dataframe. Example of dataframe
+        returned when `index_is_datetime = True`. column1     column2 index
```
gpsaggese, line 3801: You need to add a verbatim

```

to make the linter not reflow the comment.

This is a pre-existing issue, but let's fix it

**Review by gpsaggese — COMMENTED** [review]
Nice. Your code is not bad and you are picking up the code style and our idioms quickly.

Keep improving and you are getting ready

File: helpers/hpandas.py (hunk 1 above)
Comment (gpsaggese) [reply helpers/hpandas.py:764]: That's fine. Whoever wrote the code forgot to run the linter.

The best approach in these cases is to do a quick PR only running the linter so that your PR is clean.
Can you do this?

- You can create a PR SorrTask682_improve_hpandas_compare_dfs_1
with only running the linter
- Submit the PR (usually we allow to merge directly, but in this case we will do a quick review)
- Then merge master back into this PR
- Keep working on this PR

Sometimes this is called stacked PR workflow.

File: helpers/hpandas.py [hunk 9]
```diff
@@ -1960,8 +1964,13 @@ def compare_dfs(
     :param log_level: logging level
     :return: a singe dataframe with differences as values
     """
+    # Check value of `assert_diff_threshold` if exists.
```
gpsaggese, line 1967: This comment should go to line 1970

helpers/test/test_hpandas.py
```diff
@@ -2483,7 +2486,140 @@ def test8(self) -> None:
...
+        actual = hpandas.df_to_str(df_diff)
+        expected = r"""                  tsA.pct_change  tsB.pct_change  tsC.pct_change
+        timestamp
+        2022-01-01 21:01:00+00:00         0.001           0.001            0.001
+        2022-01-01 21:02:00+00:00         0.001           0.001            NaN
+        2022-01-01 21:03:00+00:00         0.001           0.001            0.001
+        """
+        self.assert_equal(actual, expected, fuzzy_match=True)
+
+    def test11(self) -> None:
```
gpsaggese, line 2549: The right approach is to do another stacked PR, where you merge the code with all the unit tests to "freeze" the behavior before changing the guts of the function.

Sometimes it's called "scaffolding the code with unit tests"

samarth9008 (commented):
TODO(Samarth): Don't merge before creating a PR on cmamp and testing it over there also

samarth9008 (changes requested):
TODO(Samarth): Don't merge before creating a PR on cmamp and testing it over there also

helpers/test/test_hpandas.py
tkpratardan: @gpsaggese 
Thank you for pointing it out. 
I still have a question about the failing unit tests of other functions of hpandas. How do you suggest we fix it, if at all?

helpers/test/test_hpandas.py
tkpratardan: GH Issue #729

helpers/hpandas.py
sonaalKant: @gpsaggese wdyt?

gpsaggese: I agree that the new code is correct.

helpers/hpandas.py
```diff
@@ -1958,10 +1962,16 @@ def compare_dfs(
...
+    :param only_warning: when `True` the function issues a warning instead of aborting
     :return: a singe dataframe with differences as values
     """
     hdbg.dassert_isinstance(df1, pd.DataFrame)
     hdbg.dassert_isinstance(df2, pd.DataFrame)
+    # Check value of `assert_diff_threshold` if exists.
+    if assert_diff_threshold:
+        hdbg.dassert_lte(assert_diff_threshold, 1.0)
+        hdbg.dassert_lte(0.0, assert_diff_threshold)
+
```
gpsaggese, line 1974: No empty line

helpers/hpandas.py
```diff
@@ -1983,47 +1993,46 @@ def compare_dfs(
...
-            "Dataframe with NaN differences=\n%s",
-            df_to_str(nan_diff_df, log_level=log_level),
-        )
-        # TODO(Grisha): add the `only_warning` switch to the function.
-        msg = "There are NaN values in one of the dataframes that are not in the other one."
-        hdbg.dassert_eq(0, nan_diff_df.shape[0], msg=msg)
     # Compute the difference df.
     if diff_mode == "diff":
-        df_diff = df1 - df2
+        try:
```
gpsaggese, line 2002: The code is correct but difficult to read.

```
# Test and convert the assertion into a boolean `is_ok`.
is_ok = True
try:
      pd.testing...
except ...
      is_ok = False
# If it's not ok, raise or not depending on `only_warning`.
if not is_ok:
     hdbd._dfatal(...)
```

Also `msg` needs to specified like "df1=%s and df2=%s don't have compatible size"

helpers/hpandas.py
```diff
@@ -1983,47 +1993,46 @@ def compare_dfs(
...
-    if remove_inf:
-        df_diff = df_diff.replace([np.inf, -np.inf], np.nan)
+            nan_mask = df_diff.isna()
+            within_threshold = (df_diff.abs() <= assert_diff_threshold) | nan_mask
+            expected = pd.DataFrame(
+                True,
+                index=within_threshold.index,
+                columns=within_threshold.columns,
+            )
+            try:
```
gpsaggese, line 2029: Same idiom as before

gpsaggese (commented):
Not bad. But we need unit tests for all the corner cases

helpers/hpandas.py
```diff
@@ -1983,47 +1993,46 @@ def compare_dfs(
...
-        # Compare NaN values in dataframes.
-        nan_diff_df = compare_nans_in_dataframes(df1, df2)
-        _LOG.log(
-            log_level,
-            "Dataframe with NaN differences=\n%s",
-            df_to_str(nan_diff_df, log_level=log_level),
-        )
-        # TODO(Grisha): add the `only_warning` switch to the function.
-        msg = "There are NaN values in one of the dataframes that are not in the other one."
-        hdbg.dassert_eq(0, nan_diff_df.shape[0], msg=msg)
```
tkpratardan, line 2000: @gpsaggese @sonaalKant 
There is one corner case here. In the context of the new code, `compare_nans` is redundant because we cannot explicitly mention the comparison of just some particular invariants. The thing is, this is to some extent mitigated by the use of `only_warning` where only a warning is issued. But the consider the corner case with `only_warning = True`:

- df1 and df2 have NaNs at different locations. (which will raise an assertion but will not abort the function with only_warning switch)

- At a particular location, the difference between two elements in df1 and df2 has exceeded the limit established by `pd.testing.assert_frame_equal()`. This will also raise an assertion.

In our case, if we want to ignore NaN comparison, but not the difference threshold, we cannot specifically do that in the new implementation. 

Could you please let me know if this behavior is valid? Or should we work it around?

helpers/hpandas.py
tkpratardan: @gpsaggese 
I understood the changes in the code structure, but I am not sure I followed the implementation of the message. Are you suggesting the we include the dataframes in the message through the custom `df_to_str()` that we built or the location/values of the elements where the comparison is failing?

helpers/hpandas.py
gpsaggese: Based on my interpretation, the problem is much simpler.
Now we use `compare_nans_in_dataframes` to decide if two dataframes are "compatible" from the nan point of view (based on some criteria that are in `compare_nans_in_dataframes`).

You just have to pass only_warning to dassert

```
        msg = "There are NaN values in one of the dataframes that are not in the other one."
        hdbg.dassert_eq(0, nan_diff_df.shape[0], msg=msg, only_warning=only_warning)
```

Makes sense?

helpers/hpandas.py
gpsaggese: Correct. When you throw an assertion you need to give as much context as possible.
E.g., imagine the code runs for 3 hours and then asserts with a message like "The dfs have not compatible size", what do you do next? Change the assertion and run again for 3 hours? How many iterations it takes to fix?

You want to give as much info as possible so you can fix the bug in one iteration.
In this case printing the size might help a bit. Sometimes we dump the dfs to disk for further analysis, but in this case it's overkilling.

Makes sense?

helpers/hpandas.py
gpsaggese: You can resolve this.

gpsaggese (commented):
Let's merge master with your 2 PRs, make the changes and merge.

We are close. This one was a bit difficult as first issue, but the work was not bad

helpers/hpandas.py
tkpratardan: Yes

gpsaggese: The issue is how to interpret the specs. I believe you are interpreting them in a more stringent that what they actually mean.

To be pragmatic:
- Let's merge the unit tests first, and then you can create a unit test for this case that you think it's problematic.
- Also it's too complicated to understand each other on GH. In these cases, we would do a quick call. We can set up an office hour to discuss all these issues needing high bandwidth.

helpers/hpandas.py
gpsaggese: Ok. If you are ok, you can make this change and resolve this.

helpers/hpandas.py
```diff
@@ -1986,47 +1993,72 @@ def compare_dfs(
...
+        # Check `is_ok` and raise an assertion depending on `only_warning`
+        if not is_ok:
+            hdbg._dfatal(
+                assertion,
+                "df1=\n%s\n and df2=\n%s\n don't have compatible size",
+                df_to_str(df1, log_level=log_level),
+                df_to_str(df2, log_level=log_level),
+                only_warning=only_warning,
+            )
+        # Calculate the difference.
```
tkpratardan, line 2018: @gpsaggese 
Let me know if I am going in the right direction, and if this is how you expect the code structure to look like

helpers/hpandas.py
tkpratardan: @gpsaggese 
Sure. Thank you

helpers/hpandas.py
```diff
@@ -1961,10 +1963,15 @@ def compare_dfs(
         if True, otherwise keep the actual result
     :param remove_inf: replace +-inf with `np.nan`
     :param log_level: logging level
+    :param only_warning: when `True` the function issues a warning instead of aborting
```
gpsaggese, line 1966: Reflow with linter

helpers/hpandas.py
```diff
@@ -1961,10 +1963,15 @@ def compare_dfs(
...
     :return: a singe dataframe with differences as values
     """
     hdbg.dassert_isinstance(df1, pd.DataFrame)
     hdbg.dassert_isinstance(df2, pd.DataFrame)
+    # Check value of `assert_diff_threshold` if exists.
```
gpsaggese, line 1971: -> , if it was passed.

helpers/hpandas.py
```diff
@@ -1986,47 +1993,72 @@ def compare_dfs(
...
-        hdbg.dassert_eq(0, nan_diff_df.shape[0], msg=msg)
     # Compute the difference df.
     if diff_mode == "diff":
+        # Test and convert the assertion into a boolean.
+        is_ok = True
+        try:
+            pd.testing.assert_frame_equal(df1, df2, check_like=True, check_dtype=False)
+        except AssertionError as e:
+            is_ok = False
+            assertion = e
```
gpsaggese, line 2008: You don't use `assertion`, so I wouldn't store it.

You can do `_ = e` to make the linter happy

helpers/hpandas.py
```diff
@@ -1986,47 +1993,72 @@ def compare_dfs(
...
+        # Check `is_ok` and raise an assertion depending on `only_warning`
```
gpsaggese, line 2009: Always `.` at the end of a comment

helpers/hpandas.py
gpsaggese: That's good.

helpers/hpandas.py
```diff
@@ -1986,47 +1993,72 @@ def compare_dfs(
...
+                    within_threshold, expected, check_exact=True
+                )
+            except AssertionError as e:
+                is_ok = False
+                assertion = e
+            # Check `is_ok` and raise assertion depending on `only_warning`.
+            if not is_ok:
+                hdbg._dfatal(
+                    assertion,
+                    "df1=\n%s\n and df2=\n%s\n don't have compatible size",
```
gpsaggese, line 2054: -> are not equal.

gpsaggese (commented):
Going in the right direction. Good work

tkpratardan:
@gpsaggese @sonaalKant 
Fixed the nits. Please verify.

helpers/hpandas.py
```diff
@@ -1986,47 +1993,74 @@ def compare_dfs(
...
+                    within_threshold, expected, check_exact=True
+                )
+            except AssertionError as e:
+                is_ok = False
+                _ = e
+            # Check `is_ok` and raise assertion depending on `only_warning`.
+            if not is_ok:
+                hdbg._dfatal(
+                    _,
+                    "df1=\n%s\n and df2=\n%s\n have pct_change more than `assert_diff_threshold`.",
```
tkpratardan, line 2056: @sonaalKant @gpsaggese 
I changed the assertion message here, which I feel gives more information. Let me know if this okay.

helpers/hpandas.py
tkpratardan: @gpsaggese 
I guess the problem would be clear now. As you can see, test8 is failing (NaN comparison) because of `only_warning = True` and no assertion is raised. However, to avoid the regression, if we setup `only_warning = False`, some of the testcases setup previously would fail because the element-wise difference between the two dfs is more than the threshold. So my point being, it would be difficult to control two types of comparisons with a single switch.

helpers/test/test_hpandas.py
```diff
@@ -3095,9 +3234,9 @@ def test1(self) -> None:
         shape=(3, 4)
                                 asset1.pct_change                asset2.pct_change
         timestamp                   high.pct_change low.pct_change   high.pct_change low.pct_change
-        2022-01-01 21:02:00+00:00        -32.881643    -287.700041        -94.505475    -259.066028
-        2022-01-01 21:03:00+00:00       -246.576815      47.525948       -137.632125      36.090517
-        2022-01-01 21:04:00+00:00       -185.862978     765.280229       -153.498432    -198.418808
+        2022-01-01 21:02:00+00:00        -32.881643     287.700041        -94.505475    -259.066028
+        2022-01-01 21:03:00+00:00        246.576815      47.525948       -137.632125      36.090517
+        2022-01-01 21:04:00+00:00        185.862978     -765.280229       -153.498432     198.418808
```
tkpratardan, line 3239: @gpsaggese @sonaalKant 
Since we change the formula for calculating the pct_change (addition of df2.abs()). I also changed the values of a test case of another function that use using compare_dfs(). Let me know if this is okay.

gpsaggese: That's correct

helpers/hpandas.py
gpsaggese: No need to tag us. It's clear from the context.

Good change.

helpers/hpandas.py
```diff
@@ -1986,47 +1993,84 @@ def compare_dfs(
...
+                pd.testing.assert_frame_equal(
+                    within_threshold, expected, check_exact=True
+                )
+            except AssertionError as e:
+                is_ok = False
+                _ = e
+            # Check `is_ok` and raise assertion depending on `only_warning`.
+            if not is_ok:
+                hdbg._dfatal(
+                    _,
```
gpsaggese, line 2065: Unclear why you pass _ here.

Better to construct a message and then pass it, in fact I think the interface is being used in a weird way

dfatal(message: str, assertion_type: Optional[Any] = None)

I'll merge if all tests are passing, and then you can fix in a different PR

tkpratardan: > dfatal(message: str, assertion_type: Optional[Any] = None)

Isn't hdbg.dfatal() used only for aborting (no `only_warning`)? I guess we need to construct if-else blocks based on the value of `only_warning` to use this. Did I understand this correctly?


> Better to construct a message and then pass it,

Sure, I will make this change.

> in fact I think the interface is being used in a weird way

I am not sure I can follow your point of view here. I used my intuition based on the guidelines to use this interface in this implementation, as I couldn't find examples of its usage. In this particular case, the assertion error was raised by `assert_frame_equal()`, and to print the error with modifications, I had to use _dfatal(). This is unlike the using hdbg function which uses _dfatal() internally 

> Unclear why you pass _ here.

Honestly, I couldn't understand the actual rationale as well. (:D)
I supposed it would mess up linter from your suggestion [here](https://github.com/sorrentum/sorrentum/pull/717#discussion_r1527632562). This whole change was to make the code more readable.

Issue #716, opened 2024-03-06
Update Sorrentum dev container


gpsaggese:
https://github.com/sorrentum/sorrentum/blob/master/sorrentum_sandbox/README.md#sorrentum-system-container

I've updated to the latest image of Airflow, built the container (instead of 2.2.2)

```
  2 #FROM apache/airflow:2.2.2
  3 FROM apache/airflow:latest
```

and docker_bash.sh doesn't work

```
$ airflow db check                                                                                                                     /home/airflow/.local/lib/python3.11/site-packages/airflow/configuration.py:812 DeprecationWarning: The sql_alchemy_conn option in [core] has been moved to the sql_alchemy_conn option in [database] - the old setting has been used, but please update your config.
/home/airflow/.local/lib/python3.11/site-packages/airflow/configuration.py:738 DeprecationWarning: The sql_alchemy_conn option in [core
] has been moved to the sql_alchemy_conn option in [database] - the old setting has been used, but please update your config.          /home/airflow/.local/lib/python3.11/site-packages/airflow/settings.py:194 DeprecationWarning: The sql_alchemy_conn option in [core] has been moved to the sql_alchemy_conn option in [database] - the old setting has been used, but please update your config.
/home/airflow/.local/lib/python3.11/site-packages/airflow/models/base.py:71 DeprecationWarning: The sql_alchemy_conn option in [core] h
as been moved to the sql_alchemy_conn option in [database] - the old setting has been used, but please update your config.
[2024-03-06T21:59:27.817+0000] {cli_action_loggers.py:135} WARNING - Failed to log action (psycopg2.OperationalError) connection to server at "host.docker.internal" (192.168.65.254), port 5532 failed: Connection refused                                                           Is the server running on that host and accepting TCP/IP connections?
...
  File "/home/airflow/.local/lib/python3.11/site-packages/sqlalchemy/engine/create.py", line 574, in connect
    return dialect.connect(*cargs, **cparams)                                                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                            File "/home/airflow/.local/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 598, in connect
    return self.dbapi.connect(*cargs, **cparams)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.11/site-packages/psycopg2/__init__.py", line 122, in connect
[...]

PR #718, opened 2024-03-07
CmTask716_update_sorrentum_dev_container

samarth9008:
#716 

- Using the latest image of airflow 
- Need to change the sequence a little bit
- After running docker-compose up, run the `./init_airflow_setup.sh` in a different terminal and then check the airflow on at 127.0.0.1:8091/home

- LMK if something else needs to be done

### DIFF
--- a/sorrentum_sandbox/devops/.env
+++ b/sorrentum_sandbox/devops/.env
@@ -3,8 +3,8 @@
 # For simplicity we use `LocalExecutor` which does not need need `Redis`
 # and `Celery`.
 # AIRFLOW__CORE__EXECUTOR=CeleryExecutor
-AIRFLOW__CORE__EXECUTOR=LocalExecutor
-AIRFLOW__CORE__SQL_ALCHEMY_CONN=postgresql+psycopg2://postgres:postgres@host.docker.internal:5532/airflow
+AIRFLOW__CORE__EXECUTOR=SequentialExecutor
+AIRFLOW__DATABSE__SQL_ALCHEMY_CONN=postgresql+psycopg2://postgres:postgres@host.docker.internal:5532/airflow
 AIRFLOW__CORE__FERNET_KEY=VYCLCTGPY5xhmKY-ReNL_RPy5F6bA7ExUO_tyScpsIo=
 AIRFLOW__CORE__LOAD_EXAMPLES=False
 AIRFLOW__WEBSERVER__SECRET_KEY=secretkey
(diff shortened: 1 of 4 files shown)


sorrentum_sandbox/devops/.env
```diff
@@ -3,8 +3,8 @@
 # For simplicity we use `LocalExecutor` which does not need need `Redis`
 # and `Celery`.
 # AIRFLOW__CORE__EXECUTOR=CeleryExecutor
-AIRFLOW__CORE__EXECUTOR=LocalExecutor
-AIRFLOW__CORE__SQL_ALCHEMY_CONN=postgresql+psycopg2://postgres:postgres@host.docker.internal:5532/airflow
+AIRFLOW__CORE__EXECUTOR=SequentialExecutor
```
samarth9008, line 6: sqlite do not work with `LocalExecutor`

sorrentum_sandbox/devops/.env
```diff
@@ -3,8 +3,8 @@
...
+AIRFLOW__DATABSE__SQL_ALCHEMY_CONN=postgresql+psycopg2://postgres:postgres@host.docker.internal:5532/airflow
```
samarth9008, line 7: SQL_ALCHEMY_CONN is part of DATABSE instead of CORE

sorrentum_sandbox/devops/docker_bash.sh
```diff
@@ -15,7 +15,7 @@ fi;
 source $DOCKER_NAME
 
 docker image ls $FULL_IMAGE_NAME
-docker manifest inspect $FULL_IMAGE_NAME | grep arch
+# docker manifest inspect $FULL_IMAGE_NAME | grep arch
```
samarth9008, line 18: Not sure but this line didn't allow me to bash into the container

sorrentum_sandbox/devops/init_airflow_setup.sh
```diff
@@ -12,7 +12,7 @@ echo "# Initializing Airflow DB setup"
 docker exec \
     -ti \
     $CONTAINER_NAME \
-    airflow db init && \
+    airflow db migrate && \
```
samarth9008, line 15: `init` is depriciated

sorrentum_sandbox/devops/docker_bash.sh
gpsaggese: It's because arch is not found in the output so grep asserts and the script aborts. we might need a `|| true` at the end, or better, find out a better way to understand what is the architecture.

Maybe something like

docker image inspect <image_name>:<tag> | grep Architecture

gpsaggese (commented):
Wonderful job!

Small nits.

Can you pls try the full example from the tutorial to make sure it works end-to-end?
I'll do the same.

PR #721, opened 2024-03-07
SorrTask707 - Add unit tests for db_utils.py.

rastislavpapco:
Only a draft of test for save_data_to_db method. Created new test class for the scenario, as it is working with "real" database, not a mock.


Er1-x:
@jsmerix When I'm running this checks, this error popped out:
```
psycopg2.OperationalError: connection to server at "127.0.0.1", port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
```
Do I need to change the config based on the server or how should I fix this ?

samarth9008:
You don't need to setup your local postgres database. It will only pass locally. Refer to the other similar test on resolving this. Provided a link on issue.

Er1-x:
> You don't need to setup your local postgres database. It will only pass locally. Refer to the other similar test on resolving this. Provided a link on issue.

I see . So we should use `helpers.hsql` instead to set up the connection, right ?

jsmerix:
> I see . So we should use helpers.hsql instead to set up the connection, right ?

The test class should work on the same principle as the example mentioned in the issue [specs](https://github.com/sorrentum/sorrentum/issues/707#issue-2163489576):

https://github.com/sorrentum/sorrentum/blob/master/im_v2/ccxt/db/test/test_archive_db_data_to_s3.py

E.g. the test class should inherit from `imvcddbut.TestImDbHelper` if it doesn't work for some reason we want to document the error in the issue and try to resolve it.

PR #753, opened 2024-03-29
 Sorr task752 add unit test db data 

Jd8111997:
@jsmerix  @sonaalKant I added a function class to test mock interaction with `load_db_data` function. 
Could you guys please check the logic for the function and let me know whether I am going in the correct direction.


im_v2/common/db/test/test_db_utils.py
```diff
@@ -9,6 +9,38 @@
...
+                                       time_interval_closed=time_interval_closed)
+        expected_query = (
+            f"SELECT * FROM {src_table} WHERE timestamp >= {start_ts.value} AND timestamp <= {end_ts.value} "
+            f"AND currency_pair IN ('BTC_USDT') AND level IN (1, 2) AND exchange_id = 'Exchange1' "
+            f"ORDER BY timestamp DESC LIMIT 100"
+        )
+        mock_execute_query_to_df.assert_called_once_with(db_connection, expected_query)
+        
+        # Assert that the function returns the expected DataFrame
+        pd.testing.assert_frame_equal(result, mock_df_result)
```
jsmerix, line 40: This part is not really necessary in my POV, you're just checking if your mock returns what you mocked.

jsmerix (changes requested):
Good direction, let's also pay attention to design the test class reusably, factor out common code as the number of test cases grow

im_v2/common/db/test/test_db_utils.py
```diff
@@ -1,4 +1,4 @@
-import unittest.mock as umock
+import unittest.mock as umock, patch
```
jsmerix, line 1: I guess this is just a draft update, but this import is incorrect - you can also see the GH action failing

im_v2/common/db/test/test_db_utils.py
```diff
@@ -9,6 +9,109 @@
...
+    # This will be run before and after each test.
+    @pytest.fixture(autouse=True)
+    def setup_teardown_test(self):
+        """
+        Initializing some common attributes for the query.
+        """
+        self.db_connection = umock.MagicMock()
+        self.src_table = "test_table"
+        self.start_ts = pd.Timestamp("2024-01-01")
+        self.end_ts = pd.Timestamp("2024-01-31")
```
jsmerix, line 22: Couple of suggestions here:
- If this is set once and not changing we don't need this to be in a fixture
- If this code was supposed to run before each test, the method name should be different + the comment above updated

im_v2/common/db/test/test_db_utils.py
```diff
@@ -9,6 +9,109 @@
...
+
+    @patch('im_v2.common.db.db_utils.hsql.execute_query_to_df')
```
jsmerix, line 24: I think this is the type of patch that would make sense to do in a fixture in order not to repeat the line in every test case.

im_v2/common/db/test/test_db_utils.py
```diff
@@ -9,6 +9,109 @@
...
+    def test_load_db_data(self, mock_execute_query_to_df):
+        """
+        Test case : Basic query construction with all the parameters provided.
```
jsmerix, line 27: Conventions:
- docstring is written in imperative
- Following an empty line you can provide additional information

jsmerix (changes requested):
Good progress.

im_v2/common/db/test/test_db_utils.py
Jd8111997: You're absolutely correct. My apologies for the oversight.

im_v2/common/db/test/test_db_utils.py
Jd8111997: @jsmerix I don't understand your first feedback. Isn't fixture is used to set up preconditions or states / attributes needed for the tests to run, such as creating test data, initializing objects, or configuring the environment ?
Additionally, in some test cases, these attributes might get modified, and thus, I re-initialize them after every test case to maintain consistency and independence between tests.

jsmerix: It's a cosmetic thing, based on the usage I observed these are more like constants rather then something that gets modified - after all why would you modify a class attribute `src_table` in a single test case rather than just creating a local variable? But not something to lose sleep over, okay to leave as is.

im_v2/common/db/test/test_db_utils.py
```diff
@@ -9,6 +9,139 @@
...
+        """
+        # Invalid timestamp format
+        invalid_ts = "invalid_timestamp"
+        # Assert that a TypeError is raised
+        with self.assertRaises(TypeError):
+            imvcddbut.load_db_data(self.db_connection, self.src_table, invalid_ts, self.end_ts)
+    
+    def test_load_db_data7(self, mock_execute_query_to_df):
+        """
+        Test case: Limit parameter is larger than total number of rows returned by database query
```
jsmerix, line 132: Let's follow the conveiton of imperative in first sentence of docstring (same for others)

jsmerix (changes requested):
One final remark, otherwise nicely done.

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
...
+        self.end_date = pd.Timestamp("2024-01-31")
+        self.start_ts = hdateti.convert_timestamp_to_unix_epoch(self.start_date, unit="ms")
+        self.end_ts = hdateti.convert_timestamp_to_unix_epoch(self.end_date, unit="ms")
+
+
+    def test_load_db_data(self):
+        """
+        Test if the query construction is done correctly when all parameters are provided.
+        """
+        # Prepare mock data
```
gpsaggese, line 31: . At the end of comments

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
...
+        """
+        Test if the function handles the scenario where the start timestamp is greater than the end timestamp.
+        """
+        # Swap start_ts and end_ts
+        start_ts, end_ts = self.end_date, self.start_date
+        # Mocking execute_query_to_df method
+        with umock.patch('im_v2.common.db.db_utils.hsql.execute_query_to_df') as mock_execute_query_to_df:
+            mock_execute_query_to_df.side_effect = ValueError
+            # Assert that function is raising ValueError
+            with self.assertRaises(ValueError):
```
gpsaggese, line 87: We need to check the value of the assertion, otherwise it's too loose.

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
...
+        """
+        super().setUp()
+        self.db_connection = umock.MagicMock()
+        self.src_table = "test_table"
+        self.start_date = pd.Timestamp("2024-01-01")
+        self.end_date = pd.Timestamp("2024-01-31")
+        self.start_ts = hdateti.convert_timestamp_to_unix_epoch(self.start_date, unit="ms")
+        self.end_ts = hdateti.convert_timestamp_to_unix_epoch(self.end_date, unit="ms")
+
+
```
gpsaggese, line 26: Linter

gpsaggese:
@DanilYachmenev can you pls nitpick the style until it's perfect

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
...
+            )
+            # Assert that an expected_query is called.
+            mock_execute_query_to_df.assert_called_once_with(self.db_connection, expected_query)
+        
+    def test_load_db_data1(self):
+        """
+        Test if the function behaves correctly when no data is returned from the database query.
+        """
+        # Mocking execute_query_to_df method
+        with umock.patch('im_v2.common.db.db_utils.hsql.execute_query_to_df') as mock_execute_query_to_df:
```
jsmerix, line 56: I don't really understand what's going on here, why have we switched to repeating code instead of mocking once?

jsmerix (changes requested):

im_v2/common/db/test/test_db_utils.py
Jd8111997: @gpsaggese, I completely agree with you on this. It is too ambiguous. However, since we are mocking the behaviors for both the `db_connection` and `execute_query_to_df` methods, how do we determine the actual `ValueError` message? As I understand it, we need to have access to the actual database to replicate the exact error message that would occur in a real execution.

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
 import pytest
 
 import helpers.hunit_test as hunitest
+import helpers.hdatetime as hdateti
```
DanilYachmenev, line 8: Always run linter on all the changed files before sending PR for review
https://github.com/sorrentum/sorrentum/blob/master/docs/coding/all.submit_code_for_review.how_to_guide.md#run-linter

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
...
+            mock_execute_query_to_df.side_effect = ValueError
+            # Assert that function is raising ValueError
+            with self.assertRaises(ValueError): 
+                result = imvcddbut.load_db_data(self.db_connection, self.src_table, start_ts, end_ts) 
+
+    def test_load_db_data4(self):
+        """
+        Test if the function raises a ValueError when invalid currency pairs are provided.
+        """
+        # Invalid currency pairs
```
DanilYachmenev, line 94: Finish all the comments with a dot`.`

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
...
+        super().setUp()
+        self.db_connection = umock.MagicMock()
+        self.src_table = "test_table"
+        self.start_date = pd.Timestamp("2024-01-01")
+        self.end_date = pd.Timestamp("2024-01-31")
+        self.start_ts = hdateti.convert_timestamp_to_unix_epoch(self.start_date, unit="ms")
+        self.end_ts = hdateti.convert_timestamp_to_unix_epoch(self.end_date, unit="ms")
+
+
+    def test_load_db_data(self):
```
DanilYachmenev, line 27: to DRY the code, just name test methods like `test1` , `test2`, etc
no need to repeat the same name all the time

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
...
+        """
+        Test if the query construction is done correctly when all parameters are provided.
+        """
+        # Prepare mock data
+        currency_pairs = ["BTC_USDT"]
+        limit = 100
+        bid_ask_levels = [1, 2]
+        exchange_id = "Exchange1"
+        time_interval_closed = True
+        # Mocking execute_query_to_df method
```
DanilYachmenev, line 37: Avoid using param names in comments whenever possible
because the name of the params are changing and this will require to keep an eye in the comment as well
unnecessary, potentially misleading

you can just say `# Mock the query.`

im_v2/common/db/test/test_db_utils.py
```diff
@@ -5,10 +5,143 @@
...
+            result = imvcddbut.load_db_data(self.db_connection, self.src_table, self.start_date, self.end_date,
+                                            currency_pairs=currency_pairs, limit=limit,
+                                            bid_ask_levels=bid_ask_levels, exchange_id=exchange_id,
+                                            time_interval_closed=time_interval_closed)
+            expected_query = (
+                f"SELECT * FROM {self.src_table} WHERE timestamp >= {self.start_ts} AND timestamp <= {self.end_ts} "
+                f"AND currency_pair IN ('BTC_USDT') AND level IN (1, 2) AND exchange_id = 'Exchange1' "
+                f"ORDER BY timestamp DESC LIMIT 100"
+            )
+            # Assert that an expected_query is called.
```
DanilYachmenev, line 48: check in all the comments
together with putting dot at the end

im_v2/common/db/test/test_db_utils.py
DanilYachmenev: @Jd8111997 see `test_load_db_table3()` in 
`im_v2/common/data/client/test/test_im_raw_data_client.py`

you need to assert the error message

DanilYachmenev:
@Jd8111997 also remember to update PR labels
pls read https://github.com/sorrentum/sorrentum/blob/master/docs/coding/all.submit_code_for_review.how_to_guide.md

im_v2/common/db/test/test_db_utils.py
Jd8111997: Thanks @DanilYachmenev, I will take a look.

im_v2/common/db/test/test_db_utils.py
jsmerix: > also tried to initialize a mock_execute_query_to_df object separately in a both setUp and tearDown method but that too didn't work. 

I cannot give you advice on this since I didn't see your code but that approach should work, it's a standard way of using mocks, examples:

https://github.com/sorrentum/sorrentum/blob/a6ea36f6cc1cbcace7640e3c7e97d0795a3f1751/im_v2/ccxt/data/extract/test/test_ccxt_extractor.py#L21

https://github.com/sorrentum/sorrentum/blob/a6ea36f6cc1cbcace7640e3c7e97d0795a3f1751/im_v2/common/data/extract/test/test_extract_utils.py#L26

Jd8111997: @jsmerix Thank you for the suggestion. It worked.

im_v2/common/db/test/test_db_utils.py
Jd8111997: @DanilYachmenev I saw the code in `im_v2/common/data/client/test/test_im_raw_data_client.py`. I still have the same concern regarding how to get the actual expected error/output precisely in a case when we are trying to mock the method and not running on actual databases.

im_v2/common/db/test/test_db_utils.py
```diff
@@ -1,15 +1,202 @@
...
+        yield
+        # Run after each test.
+        self.tear_down_test()
+
+    def set_up_test(self) -> None:
+        # Create new mocks from patch's start() method.
+        self.query_mock: umock.MagicMock = self.mock_execute_query_df.start()
+
+    def tear_down_test(self) -> None:
+        self.mock_execute_query_df.stop()
```
jsmerix, line 50: Any particular reason why you used both? Wouldn't it be simpler to only use `set_up_test`

im_v2/common/db/test/test_db_utils.py
```diff
@@ -1,15 +1,202 @@
...
+        limit = 1000000
+        # Assert that a ValueError is raised.
+        with self.assertRaises(ValueError):
+            imvcddbut.load_db_data(
+                self.db_connection,
+                self.src_table,
+                self.start_date,
+                self.end_date,
+                limit=limit,
+            )
```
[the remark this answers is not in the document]
gpsaggese: I've added a note to the unit test documentation

### Always explain self.assertRaises

- Testing for an assertion needs always to be done with the following idiom
  to explain exactly what we are catching and why
  ```python
  with self.assertRaises(AssertionError) as cm:
      hlitagit.git_patch_create(
          ctx, mode, modified, branch, last_commit, files
      )
  act = str(cm.exception)
  exp = r"""
  * Failed assertion *
  '0'
  ==
  '1'
  Specify only one among --modified, --branch, --last-commit
  """
  self.assert_equal(act, exp, fuzzy_match=True)
  ``

im_v2/common/db/test/test_db_utils.py
Jd8111997: Fixed it!

im_v2/common/db/test/test_db_utils.py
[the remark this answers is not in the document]
Jd8111997: @gpsaggese @jsmerix I have a question. In the example you provided, you've used the `exp` variable to handle the exception more precisely. My question is, how can we retrieve the exact exp message for the method we're attempting to mock? Since we're not directly executing queries on the actual database, how can we determine what the exact error message would be?"

In our example, we are mocking both `db_connection` object and `execute_query_to_df` method.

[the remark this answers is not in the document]
gpsaggese: IIUC, you should put whatever assertion is throwing.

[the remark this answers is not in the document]
Jd8111997: Thanks @gpsaggese.

im_v2/common/db/test/test_db_utils.py
```diff
@@ -1,15 +1,181 @@
...
+        Test if the query construction is done correctly when all parameters
+        are provided.
+        """
+        # Prepare mock data.
+        currency_pairs = ["BTC_USDT"]
+        limit = 100
+        bid_ask_levels = [1, 2]
+        exchange_id = "Exchange1"
+        time_interval_closed = True
+
```
gpsaggese, line 57: No empty space.

You want to say
```
# Prepare data.
...
# Run test.
...
# Check results.
```

im_v2/common/db/test/test_db_utils.py
```diff
@@ -1,15 +1,181 @@
...
+            self.db_connection,
+            self.src_table,
+            self.start_timestamp,
+            self.end_timestamp,
+        )
+        expected_query = f"SELECT * FROM {self.src_table} WHERE timestamp >= {self.start_ts} AND timestamp <= {self.end_ts}"
+        self.query_mock.assert_called_once_with(
+            self.db_connection, expected_query
+        )
+        # Assert that function is returning None.
```
gpsaggese, line 96: The code is clear enough. No need for comment.

im_v2/common/db/test/test_db_utils.py
```diff
@@ -1,15 +1,181 @@
...
+        """
+        # Mock the query.
+        self.query_mock.return_value = None
+        result = imvcddbut.load_db_data(
+            self.db_connection,
+            self.src_table,
+            self.start_timestamp,
+            self.end_timestamp,
+        )
+        expected_query = f"SELECT * FROM {self.src_table} WHERE timestamp >= {self.start_ts} AND timestamp <= {self.end_ts}"
```
gpsaggese, line 92:
```
# Check results.
```
goes here

gpsaggese (commented):
Minor tweaks and then let's merge

Issue #763, opened 2024-04-04
Comment out Crypto chasis extractor


sonaalKant:
Ref. https://github.com/cryptokaizen/cmamp/issues/7778

- Comment out chasis extractor: https://github.com/sorrentum/sorrentum/blob/38811f93d51f0a4b176ccba7eca0938f5c2f55d9/im_v2/crypto_chassis/data/extract/extractor.py#L4
- Skip all the related tests.

PR #767, opened 2024-04-05
Comment out Crypto chasis extractor

Vedanshu7:
#763 
- Comment out the code
- Skipped the related test cases


im_v2/crypto_chassis/data/extract/test/test_chassis_extractor.py
```diff
@@ -11,33 +11,34 @@
 _LOG = logging.getLogger(__name__)
 
 
+@pytest.mark.skip(reason="cmamp #7778.")
 class TestCryptoChassisExtractor1(hunitest.TestCase):
-    pandas_read_csv_patch = umock.patch.object(
```
sonaalKant, line 15: Does commenting out stuff necessary, can't we just add a skip ?

sonaalKant (approved):

im_v2/crypto_chassis/data/extract/test/test_chassis_extractor.py
Vedanshu7: @sonaalKant 
If I don't comment the code, an error will occur. The skip method for the class will then skip the execution of the test cases, but the class will still be constructed. To address this, I suggest setting the initial value of the parameter to None and creating an initialize method. Afterward, calling the initialize method within the set_up_test method will initialize these parameters. Let me know if it's a fine solution.

Vedanshu7: Resolved.

PR #769, opened 2024-04-06
Add new code to test_hintrospection.Test_is_pickle

ttiaNlluu:
@DanilYachmenev able1: unit test not callable input(int or str)


gpsaggese:
You need to add @DanilYachmenev as reviewer

ttiaNlluu:
Thanks for your reminder, I just added @DanilYachmenev as reviewer.

helpers/test/test_hintrospection.py
```diff
@@ -30,6 +30,24 @@ def say(self) -> None:
...
+class _ClassNonPickleable:
+    """
+    Class with non-pickleable param values.
+    """
+
+    def __init__(self) -> None:
+        self._arg1 = lambda x: x
+        self._arg2 = 2
+
 class Test_is_pickleable1(hunitest.TestCase):
```
DanilYachmenev, line 51: Good finding but lets not change this test class for now
Especially since it it checking many other functions and an actually object pickleability, not directly `is_pickleable()`

helpers/test/test_hintrospection.py
```diff
@@ -66,6 +84,11 @@ def helper(
         _LOG.debug("act_pickled=%s", act_pickled)
         _LOG.debug("exp_pickled=%s", exp_pickled)
         self.assertEqual(act_pickled, exp_pickled)
+        #type_search to pickle
+        act_pickled = hintros.is_pickleable(obj)
+        _LOG.debug("act_pickled=%s", act_pickled)
+        _LOG.debug("exp_pickled=%s", exp_pickled)
+        self.assertEqual(act_pickled, exp_pickled)
```
DanilYachmenev, line 91: so extending this test class will only increase test complexity and vulnerability

Rule - 1 test class checks 1 function
Therefore, create `Test_is_pickleable2` and simply test it with different input types

helpers/test/test_hintrospection.py
DanilYachmenev, line 51: for `Test_is_pickleable2` add a docstring like

```
Check that `is_pickleable()` works correctly.
```

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +183,49 @@ def test_method4(self) -> None:
...
+        # Not callable: str.
+        str_ = "hello"
+        func = str_
+        exp_str = r"hello"
+        # A str is bound to an object.
+        exp_bound = True
+        exp_lambda = False
+        # A str is pickleable.
+        exp_pickled = True
+        self.helper(func, exp_str, exp_bound, exp_lambda, exp_pickled)
```
DanilYachmenev, line 208: lets not test ints and strings in separate test methods - overkill
use smth like `[1, "2", 0.3]` as inout for 1 test to check widespread non-callables right away
or just do for 1 type

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +183,49 @@ def test_method4(self) -> None:
         exp_pickled = True
         self.helper(func, exp_str, exp_bound, exp_lambda, exp_pickled)
 
+    def test_int(self) -> None:
```
DanilYachmenev, line 186: all test method names should end with a digit like -> `test_int1`

However, not sure if it is the right method name rn
lets do `test_non_callable1`

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +183,49 @@ def test_method4(self) -> None:
...
+        self.helper(func, exp_str, exp_bound, exp_lambda, exp_pickled)
+
+    def test_class1(self) -> None:
+        # A class object with pickleable param values.
+        class_instance = _ClassPickleable()
+        func = class_instance
+        exp_str = r"<__main__._ClassPickleable object at 0x104786bb0>"
+        exp_bound = True
+        exp_lambda = False
+        exp_pickled = True
```
DanilYachmenev, line 217: so no need to check all of it in a new test class
just the tested function outcomes or assertion error messages

DanilYachmenev (commented):
good progress, good catch of the test method
however, lets do everything clearer and separate
see comments below

ttiaNlluu:
@DanilYachmenev create Test_is_pickleable2 to test is_pickleable()

DanilYachmenev:
> @DanilYachmenev create Test_is_pickleable2 to test is_pickleable()

yes, see https://github.com/sorrentum/sorrentum/pull/769#discussion_r1558157333

`Test_is_pickleable1` is not testing something else but not the function we need
let it be as it is and create another one

ttiaNlluu:
> > @DanilYachmenev create Test_is_pickleable2 to test is_pickleable()
> 
> yes, see [#769 (comment)](https://github.com/sorrentum/sorrentum/pull/769#discussion_r1558157333)
> 
> `Test_is_pickleable1` is not testing something else but not the function we need let it be as it is and create another one

Yes, I have created `class Test_is_pickleable2` to only test `is_pickleable()`, and return `class Test_is_pickleable1` to its original version. Please take a review.

DanilYachmenev:
> Yes, I have created `class Test_is_pickleable2` to only test `is_pickleable()`, and return `class Test_is_pickleable1` to its original version. Please take a review.

Please, follow our PR process
no need to ping reviewers in PR comments for example
read https://github.com/sorrentum/sorrentum/blob/master/docs/coding/all.submit_code_for_review.how_to_guide.md#save-reviewer-time

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +159,123 @@ def test_method4(self) -> None:
...
+        self._arg2 = ["2", 3]
+
+class _ClassNonPickleable:
+    """
+    Class with non-pickleable param values.
+    """
+
+    def __init__(self) -> None:
+        self._arg1 = lambda x: x
+        self._arg2 = 2
```
DanilYachmenev, line 178: Lets move it up to the beginning of section for `Test_is_pickleable` test classes

And lets merge  `_ClassPickleable`  and `Hello` class together for DRY reasons
so we will have the following test classes:

```
class _ClassPickleable:
    """
    Class with pickleable param values.
    """

    def __init__(self) -> None:
        self._arg1 = 1
        self._arg2 = ["2", 3]

    @staticmethod
    def say2(self) -> None:
        print("Hello")

    def say(self) -> None:
        print("Hello")


class _ClassNonPickleable:
    """
    Class with non-pickleable param values.
    """

    def __init__(self) -> None:
        self._arg1 = lambda x: x
        self._arg2 = 2
```
And update the calls in `Test_is_pickleable1`

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +159,123 @@ def test_method4(self) -> None:
...
+    """
+    Check that `is_pickleable()` works correctly.
+    """
+    def helper(
+        self,
+        obj: Any,
+        exp_pickled: bool,
+    ) -> None:
+        _LOG.debug("obj=%s", obj)
+        #type_search to pickle
```
DanilYachmenev, line 190: lets remove this comment and add a docstring to the method like
```
"""
Run the function and verify the outcome.
"""
```

code style of `Test_is_pickleable1` is a bit obsolete with our current one - we no add doctstring to every method (unless its self-evident)

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +159,123 @@ def test_method4(self) -> None:
...
+        self.assertEqual(act_pickled, exp_pickled)
+
+    def test_non_callable1(self) -> None:
+        # Not callable (int, str, float)
+        non_callable = [1, "2", 0.3]
+        for i in non_callable:
+            func = non_callable
+            # Not callable objects are pickleable.
+            exp_pickled = True
+            self.helper(func, exp_pickled)
```
DanilYachmenev, line 203: lets avoid running cycles in test methods and instead split it on 2 methods:
- one has `1` as input
- another has a list `[1, "2", 0.3]`

this will cover 2 cases with non-callables (non-callable value and an Iterable) which is sufficient, no need to test all the possible non-callable types here

add docstrings like 
```
"""
Check that a non-callable is pickleable.
"""
```
and
```
Check that a non-callable Iterable is pickleable.
```

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +159,123 @@ def test_method4(self) -> None:
...
+        # A lambda is not pickleable.
+        exp_pickled = False
+        self.helper(func, exp_pickled)
+
+    def test_lambda2(self) -> None:
+        lambda_ = lambda x: x
+        func = lambda_
+        # A lambda is not pickleable.
+        exp_pickled = False
+        self.helper(func, exp_pickled)
```
DanilYachmenev, line 218: lets keep only this lambda test, no need for the one above

add a docstring is well
```
"""
Check that a lambda function is not pickleable.
"""
```

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +159,123 @@ def test_method4(self) -> None:
...
+
+    def test_func1(self) -> None:
+        def _hello() -> bool:
+            return False
+        #
+        func = _hello
+        # A local object is not pickleable.
```
DanilYachmenev, line 225: move to a docstring like 

```
"""
Check that a local object is not pickleable. 
"""
```

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +159,123 @@ def test_method4(self) -> None:
...
+    def test_func2(self) -> None:
+        # Global function.
+        func = hello
+        # A global function is pickleable since it's not bound locally or
+        # to an object.
+        exp_pickled = True
+        self.helper(func, exp_pickled)
+
+    def test_method1(self) -> None:
+        # A class method but unbound to an object.
```
DanilYachmenev, line 238: Basically try to convert such comments in all the test methods to the docstrings in a similar way I suggest ablove

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +159,123 @@ def test_method4(self) -> None:
...
+
+    def test_class1(self) -> None:
+        # A class object with pickleable param values.
+        class_instance = _ClassPickleable()
+        func = class_instance
+        exp_pickled = True
+        self.helper(func, exp_pickled)
+
+    def test_class2(self) -> None:
+        # A class object with non-pickleable param values.
```
DanilYachmenev, line 273: here lets add a docstring like
```
Check that a class with not pickleable param values is considered pickleable.
```

this is a wrong behaviour that is being fixed in an ongoing issue
I will ping you when it is merged and integrated to sorrentum to extend the coverage

DanilYachmenev (commented):
nice coverage approach
a bit too much but it is better than less

see the comments below, beware the code style

helpers/test/test_hintrospection.py
```diff
@@ -29,6 +29,31 @@ def say2(self) -> None:
     def say(self) -> None:
         print("Hello")
 
+class _ClassPickleable:
```
DanilYachmenev, line 32: Deprecate `Hello` class and use `_ClassPickleable` instead in all the tests

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +185,122 @@ def test_method4(self) -> None:
...
+    """
+    Test that `is_pickleable()` works correctly.
+    """
+    def helper(
+        self,
+        obj: Any,
+        exp_pickled: bool,
+    ) -> None:
+        """
+        Run is_pickleable() function and verify the outcome.
```
DanilYachmenev, line 198: -> `Run the function and verify the outcome.`

In general, avoid using object names in comments and docstrings
names are changing often and then we should keep in mind updating all their mentions in the comments - we want to avoid this

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +185,122 @@ def test_method4(self) -> None:
         exp_pickled = True
         self.helper(func, exp_str, exp_bound, exp_lambda, exp_pickled)
 
+class Test_is_pickleable2(hunitest.TestCase):
+    """
+    Test that `is_pickleable()` works correctly.
+    """
```
DanilYachmenev, line 191: no need for a docstring here

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +185,122 @@ def test_method4(self) -> None:
...
+        """
+        Check that a global object is pickleable. 
+        """
+        func = hello
+        exp_pickled = True
+        self.helper(func, exp_pickled)
+
+    def test_method1(self) -> None:
+        """
+        Check that a unbound class method is pickleable.
```
DanilYachmenev, line 254: micro nit
-> `Check that an ...`

helpers/test/test_hintrospection.py
```diff
@@ -160,6 +185,122 @@ def test_method4(self) -> None:
...
+        exp_pickled = True
+        self.helper(func, exp_pickled)
+
+    def test_method3(self) -> None:
+        """
+        Check that a bound method is pickleable. 
+        """
+        class_instance = _ClassPickleable()
+        func = class_instance.say
+        exp_pickled = True
```
DanilYachmenev, line 274: Shouldn't it be not pickleable?

what about these lines in the tested func
```
    is_bound = is_bound_to_object(obj)
    _LOG.debug("is_bound=%s", is_bound)
    if is_bound:
        return False
```

helpers/test/test_hintrospection.py
ttiaNlluu: `class Test_is_pickleable1` is still using `Hello` class. Should I modify that also, and delete the `Hello` class?

helpers/test/test_hintrospection.py
ttiaNlluu: Yes, a bound method is not pickleable. I have modified `exp_pickled` value to False.

DanilYachmenev: and yet, it is still `True` in the latest PR  commit
update
and before every review run the written tests to make sure they display the expected behaviour

helpers/test/test_hintrospection.py
ttiaNlluu: `class Test_is_pickleable1` was not generated by me, so I only deprecate `Hello` class and use `_ClassPickleable` instead, I keep other codes same. Let me know if I need to further modification.

ttiaNlluu:
Sorry about that, I have completed it on Monday. Now I understand that I need to ask for review again.

helpers/test/test_hintrospection.py
```diff
@@ -2,6 +2,10 @@
 import os
 from typing import Any, Callable
 
+import sys
+sys.path.append('/Users/tianlulu/src/sorrentum1/')
+print(sys.path)
+
```
DanilYachmenev, line 8: why is it here?

helpers/test/test_hintrospection.py
```diff
@@ -17,11 +21,18 @@
 # #############################################################################
 
 
-def hello() -> bool:
+def class_pickleable() -> bool:
```
DanilYachmenev, line 24: no need to rename this function, the discussion was only about test classes

helpers/test/test_hintrospection.py
```diff
@@ -152,14 +172,127 @@ def test_method3(self) -> None:
...
         # A static class method.
-        hello_ = Hello()
-        func = hello_.say2
+        class_instance = _ClassPickleable()
+        func = class_instance.say2
         exp_str = r"<function Hello.say2 at 0x>"
         exp_bound = False
         exp_lambda = False
         exp_pickled = True
         self.helper(func, exp_str, exp_bound, exp_lambda, exp_pickled)
```
DanilYachmenev, line 182: run linter
distance between classes is 2 lines
linter takes care of it automatically

helpers/test/test_hintrospection.py
```diff
@@ -152,14 +172,127 @@ def test_method3(self) -> None:
...
+        _LOG.debug("act_pickled=%s", act_pickled)
+        _LOG.debug("exp_pickled=%s", exp_pickled)
+        self.assertEqual(act_pickled, exp_pickled)
+
+    def test_non_callable1(self) -> None:
+        """
+        Check that a non-callable is pickleable.
+        """
+        non_callable = 1
+        func = non_callable
```
DanilYachmenev, line 203: just do 
`obj = 1`
no need to multiply lines here + giving a name that is similar to param name in `helper` would simplify understanding for those who read it for the 1st time

helpers/test/test_hintrospection.py
```diff
@@ -152,14 +172,127 @@ def test_method3(self) -> None:
...
+        exp_pickled = True
+        self.helper(func, exp_pickled)
+
+    def test_non_callable2(self) -> None:
+        """
+        Check that a non-callable Iterable is pickleable.
+        """
+        non_callable = [1, "2", 0.3]
+        func = non_callable
```
DanilYachmenev, line 212: same here and below

DanilYachmenev (commented):
also pls run the tests and post GH logs that say that everything is passing

ttiaNlluu:
Can I close this PR and open a new one? I created a branch under my forked repo, then linter failed with AssertionError.
I want to create a new branch from sorrentum/sorrentum, and re-open a PR on that.

DanilYachmenev (commented):
do what is necessary to make the tests (like making a new branch in a correct repo)
apart from it lgtm

Issue #762, opened 2024-04-04
Start downloading data 


sonaalKant:
We would like to have a pipeline to fetch data from these sources on which later we will evaluate our model. 

[goperigon.com](https://goperigon.com/)
[newsdata.io](https://newsdata.io/)
[mediastack.com](https://mediastack.com/)
[thenewsapi.com](https://thenewsapi.com/)
[marketaux.com](https://marketaux.com/)

Lets try to write the code which follows OOP principles and well documented. Let's start with using free version of each api. If needed we will upgrade.

FYI @gpsaggese

tkpratardan:
@sonaalKant 
What would be the path for the script?

gpsaggese:
What's the plan exactly? Which data sources? What is the format? What does the data look like?

Let's put together a notebook showing a bit of data so we can make an informed decision, and use the Gdoc to track the results

tkpratardan:
@gpsaggese 

> What's the plan exactly? Which data sources? What is the format? What does the data look like?

As mentioned, the plan here is to create a data pipeline to fetch data from the specified datasources. We would initially be using their free API keys (limited requests). The code would account for the following:
- Separate classes for data pipelines of each data source. These classes would take in all inputs needed for fetching the data, viz:
   1.   `endpoints`: If they exist, we will have an argument where we can specify the interested endpoint of API.
   2.  query parameters like `keywords (q), categories, topics, start date, end date` 

We get a JSON response typically consisting the article information like headline, article source, URL, article text, summary, etc. Some of the data sources provide a sentiment analysis. All articles from the API request query are then stored in a dictionary, which will be the return of the `fetch_data()` method for each class. 


> Let's put together a notebook showing a bit of data so we can make an informed decision, and use the Gdoc to track the results

I will raise a draft PR once the filepath is confirmed and update the GDoc accordingly.

Let me know if this makes sense.

sonaalKant:
Initial steps on PEOE, closing as experiment is completed and merged.

PR #771, opened 2024-04-08
SorrTask762_start_downloading_data

tkpratardan:
@sonaalKant @gpsaggese 
I have developed an initial version of the code to download data from the specified sources mention in #762. However, we have a few bottlenecks with the free API:

1. [goperigon.com](https://goperigon.com/) -> limited API request (150)
2. [newsdata.io](https://newsdata.io/) ->  daily limit of 10 articles
3. [mediastack.com](https://mediastack.com/) -> I haven't implemented a pipeline for this datasource because the free 
                                                                                        version requires payment info.
4. [thenewsapi.com](https://thenewsapi.com/) -> limited API requests per day and some access to some endpoints is 
                                                                                     blocked. Example, crypto.
5. [marketaux.com](https://marketaux.com/) -> limited API requests and access to some endpoints is blocked.

Review:
1. Is this according to expectations?
2. Do we need to implement data formatting and processing? LIke converting JSON responses to a dataframe and cleaning the texts? or is the data processing implemented in another script?


sonaalKant:
Good start, Lets get some example data points from each API and log them in some doc so that we quickly see and analyse the next steps.

research_llm/data_download.py
```diff
@@ -0,0 +1,374 @@
...
+import json
+
+_LOG = logging.getLogger(__name__)
+
+
+# %% [markdown]
+# ## goperigon.com pipeline
+
+# %%
+class goperigonPipeline:
```
samarth9008, line 38: No need for OOPs programming inside notebooks. Over here, you can try interacting with API directly. Download the data and get a sense about it.

samarth9008: But lets follow what @sonaalKant says. This is just my opinion

tkpratardan:
@gpsaggese @sonaalKant @samarth9008 
I have made changes to the notebook, including some preliminary preprocessing and text cleaning for better readability.
Since the JSON responses were nested, I unpacked them by keeping the hierarchy intact. For ex:
Screenshot from 2024-04-11 07-31-41


We can interpret this as different `keys` 
under `keywords` and same index of the list across the columns form the corresponding values to the `keys`.

research_llm/data_download.py
tkpratardan: I kept the OOP method because I feel like this way gives us more functionality through controlling attributes of class instances. This could come quite handy while extracting data with different params/endpoints, making it faster, IMO. Let me know if this is okay.

PR #885, opened 2024-04-28
Dockneo files after soft reset

sivaniDronamraju:
Used data.csv to build a graph in Neo4j. Ran cypher queries for finding Influential users, friend of friends and common connections.
Was not able to push data.csv because the file size exceeded limit. How do I proceed with this?
Built the container using docker-compose up --build but couldn't stop it with docker-compose down or stop.


Shaunak01:
You can reduce the file size or split the files into multiple files. Try to stop the container forcefully, though this should not be the ideal case. Debug as to why the problem occurred

Issue #904, opened 2024-05-02
Unit test `find_min_max_timestamps_from_intervals()`


DanilYachmenev:
There is a `find_min_max_timestamps_from_intervals()` in `dataflow/core/utils.py` that is being used in `HistoricalDataSource` 

We need to add unit tests to `dataflow/core/test/test_utils.py`.

Implement full coverage of the function but not too much.
I'd expect 5-7 small unit test methods that cover each if-else clause in the function body.

For in each method, try different value types - for this please understand `Intervals` value type beforehand ([see](https://github.com/kaizen-ai/kaizenflow/blob/f665ad177b5d3ab8bf57b8f0e0290e13d719a72f/dataflow/core/utils.py#L51))

See [unit test doc](https://github.com/sorrentum/sorrentum/blob/master/docs/coding/all.write_unit_tests.how_to_guide.md) to follow the code style

Assigning @samarth9008 to re-distribute
Could be a good 2nd task for those who already understand our approach to unit testing, here some thinking is required

FYI @gpsaggese @DanilYachmenev

PR #911, opened 2024-05-05
SorrTask904 Unit test find_min_max_timestamps_from_intervals()

neha2801-create:
#904 
 - Added Unit tests for method find_min_max_timestamps_from_intervals()

### DIFF
--- a/dataflow/core/test/test_utils.py
+++ b/dataflow/core/test/test_utils.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 from typing import Tuple
 
@@ -149,7 +150,9 @@ class Test_convert_to_multiindex(hunitest.TestCase):
         1 <= 0
         ################################################################################
         """
-        self.assert_equal(actual_error_message, expected_error_message, fuzzy_match=True)
+        self.assert_equal(
+            actual_error_message, expected_error_message, fuzzy_match=True
+        )
 
     def test4(self) -> None:
         """
@@ -186,7 +189,9 @@ class Test_convert_to_multiindex(hunitest.TestCase):
                                                       names=[None, 'id']))' is '<class 'tuple'>' instead of '<class 'pandas.core.frame.DataFrame'>'
         ################################################################################
         """
-        self.assert_equal(actual_error_message, expected_error_message, fuzzy_match=True)
+        self.assert_equal(
+            actual_error_message, expected_error_message, fuzzy_match=True
+        )
 
     def test6(self) -> None:
         """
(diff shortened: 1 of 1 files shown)


dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,181 @@ def test6(self) -> None:
...
+
+        if method_name == "empty_intervals":
+            return (None, (None, None))
+        elif method_name == "single_interval_with_none_endpoints":
+            return ([(None, None)], (None, None))
+        elif method_name == "single_interval_with_datetime_endpoints":
+            return (
+                [(COMMON_DATETIME_START, COMMON_DATETIME_END)],
+                (COMMON_DATETIME_START, COMMON_DATETIME_END),
+            )
```
DanilYachmenev, line 248: I like the coverage you do but the approach should be totally refactored
I see that you aimed to reduce the number of repeating values but this is not always the goal, especially in unit tests

However , the first and foremost priority should be a clarity of code
e.g. imagine someone who has no context reading your code - it should be clear what it does and how it is done right away
It is especially important for unit tests because it is supposed to help anyone who debugs the code to comprehend everything quickly

An imagine reading your code in this context)
while it does the work and covers most cases, it is hard to get right away

e.g. if `test2` is broken I need to go and see what is `single_interval_with_none_endpoints` switch, what are the meanings and values of set params etc
Unit test should be so clear that a bird outside of your window should understand where the bug hides

This also applies to many further tasks btw
While DRY is important, we should always prioritize clarity since it will later save not only your time but all the people here after you

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,181 @@ def test6(self) -> None:
...
+        elif method_name == "single_interval_with_same_datetime_endpoints":
+            return (
+                [(COMMON_DATETIME_START, COMMON_DATETIME_START)],
+                (COMMON_DATETIME_START, COMMON_DATETIME_START),
+            )
+        elif method_name == "single_interval_with_timestamp_endpoints":
+            return (
+                [(COMMON_TIMESTAMP_START, COMMON_TIMESTAMP_END)],
+                (COMMON_TIMESTAMP_START, COMMON_TIMESTAMP_END),
+            )
```
DanilYachmenev, line 258: This is why do not set this clauses, do not set global params
for every method just set them directly:
```
start_ts = `pd.Timestamp...`
end_ts = ...
```

and then give a short 1-liner in test docstrings like `Check for a single interval with timestamp endpoints.`

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,181 @@ def test6(self) -> None:
         actual_df = dtfcorutil.convert_to_multiindex(df, asset_id_col)
         # Compare the result.
         self.assert_multiindex_columns_equal(expected_df_columns, actual_df)
+
+
+class TestFindMinMaxTimestampsFromIntervals(hunitest.TestCase):
+    @staticmethod
+    def get_data(method_name: str) -> Tuple:
+        COMMON_YEAR = 2022
```
DanilYachmenev, line 214: do capitalise param names inside any object
caps is only for global params that are set on top of the file

I set like 5-6 during my 3 years here so in general just not use it

DanilYachmenev (commented):
good coverage but needs to be restructured first

dataflow/core/test/test_utils.py
neha2801-create: Hey @DanilYachmenev , 
Just for confirmation, I have to remove clauses like : method_name ="single_interval_with_same_datetime_endpoints"
as shown above and use other way to match condition and each test case.
May I use dictionary and match the key for different test case  or declare respective test case input values in test case body itself.

DanilYachmenev: @neha2801-create almost
In fact, you should not have such helper at all, no `if-else` clauses
make each test case almost independently, set intervals right inside
e.g.

```
def test7(self) -> None:
        """
        ...
        """
        interval1 = (pd.Timestamp(...), None)
        interval 2 = (None, datetime.datetime(...)
        intervals = ([...   ...])
        actual = dtfcorutil.find_min_max_timestamps_from_intervals(
            intervals
        )
        expected = ...
        self.assertEqual(actual, expected)
```
so no need for helpers here, set necessary inputs for each test methods inside

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+
+
+class TestFindMinMaxTimestampsFromIntervals(hunitest.TestCase):
+    def test1(self) -> None:
+        """
+        Check for the case when intervals is None.
+        """
+        intervals = None
+        expected_min_max = (None, None)
+
```
DanilYachmenev, line 218: Do not leave empty lines
fill them with comments or remove

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+        intervals = [
+            (
+                datetime.datetime(2022, 1, 1, 10, 0, 0),
+                datetime.datetime(2022, 1, 1, 11, 0, 0),
+            )
+        ]
+        expected_min_max = (
+            datetime.datetime(2022, 1, 1, 10, 0, 0),
+            datetime.datetime(2022, 1, 1, 11, 0, 0),
+        )
```
DanilYachmenev, line 246: this one is clear and nice, I'd make all the following ones in similar fashion - just put values directly

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+        intervals = [(None, None)]
+        expected_min_max = (None, None)
+
+        actual = dtfcorutil.find_min_max_timestamps_from_intervals(intervals)
+        self.assertEqual(expected_min_max, actual)
+
+    def test3(self) -> None:
+        """
+        Check for a single interval with datetime.datetime objects as
+        endpoints.
```
DanilYachmenev, line 235: -> `Check for a single interval with valid endpoints.`

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+            datetime.datetime(2022, 1, 1, 10, 0, 0),
+        )
+
+        actual = dtfcorutil.find_min_max_timestamps_from_intervals(intervals)
+        self.assertEqual(expected_min_max, actual)
+
+    def test5(self) -> None:
+        """
+        Check for a single interval with pd.Timestamp objects as endpoints.
+        """
```
DanilYachmenev, line 273: no need to repeat the same test but for timestamp now
you can just use it in another test that checks other case but use timestamps there - if it works then code works for timestamps

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+            datetime.datetime(2022, 1, 1, 11, 0, 0),
+        )
+
+        actual = dtfcorutil.find_min_max_timestamps_from_intervals(intervals)
+        self.assertEqual(expected_min_max, actual)
+
+    def test4(self) -> None:
+        """
+        Check for a single interval with the same datetime.datetime objects as
+        endpoints.
```
DanilYachmenev, line 254: do not write types of values in dostrings
you pass Interval type values that can be datetime, timestamp
but it is needed to be checked explicitly for the Interval type itself, not here

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+        intervals = [
+            (
+                pd.Timestamp(2022, 1, 1, 10, tz="UTC"),
+                pd.Timestamp(2022, 1, 1, 10, tz="UTC"),
+            )
+        ]
+        expected_min_max = (
+            pd.Timestamp(2022, 1, 1, 10, tz="UTC"),
+            pd.Timestamp(2022, 1, 1, 10, tz="UTC"),
+        )
```
DanilYachmenev, line 302: too many checks for single intervals

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+    def test7(self) -> None:
+        """
+        Check for multiple intervals with a mix of datetime.datetime,
+        pd.Timestamp objects, and None as endpoints.
+        """
+        intervals = [
+            (datetime.datetime(2022, 1, 1, 10, 0, 0), None),
+            (None, pd.Timestamp(2022, 1, 1, 11, tz="UTC")),
+        ]
+        expected_min_max = (None, None)
```
DanilYachmenev, line 316: just call `expected`
here and everywhere else

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+
+    def test7(self) -> None:
+        """
+        Check for multiple intervals with a mix of datetime.datetime,
+        pd.Timestamp objects, and None as endpoints.
+        """
+        intervals = [
+            (datetime.datetime(2022, 1, 1, 10, 0, 0), None),
+            (None, pd.Timestamp(2022, 1, 1, 11, tz="UTC")),
+        ]
```
DanilYachmenev, line 315: set intervals like this, just directly

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+            pd.Timestamp(2022, 1, 1, 10, tz="UTC"),
+        )
+
+        actual = dtfcorutil.find_min_max_timestamps_from_intervals(intervals)
+        self.assertEqual(expected_min_max, actual)
+
+    def test7(self) -> None:
+        """
+        Check for multiple intervals with a mix of datetime.datetime,
+        pd.Timestamp objects, and None as endpoints.
```
DanilYachmenev, line 310: -> `Check for multiple intervals with a mix of valid and empty endpoints.`

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+        interval2_end = interval1_end - datetime.timedelta(hours=2)
+        intervals = [
+            (interval1_start, interval1_end),
+            (interval2_start, interval2_end),
+        ]
+        overall_start = interval2_start
+        overall_end = interval1_end
+        expected_min_max = (overall_start, overall_end)
+
+        actual = dtfcorutil.find_min_max_timestamps_from_intervals(intervals)
```
DanilYachmenev, line 338: do not specify so many params
for a function it is ok but in the test - values should be visible in order to check them just by looking
so
```
intervals = [
    (datetime.datetime(2022, 1, 1, 10, 0, 0), datetime.datetime(2022, 1, 1, 14, 0, 0))
    ...
]
actual = ...
expected = ...
self.assertEqual(actual, expected)
```

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+        self.assertEqual(expected_min_max, actual)
```
DanilYachmenev, line 339: put `actual` first

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +206,153 @@ def test6(self) -> None:
...
+
+    def test8(self) -> None:
+        """
+        Check for multiple intervals with datetime.datetime objects as
+        endpoints.
+        """
+        interval1_start = datetime.datetime(2022, 1, 1, 10, 0, 0)
+        interval1_end = datetime.datetime(2022, 1, 1, 11, 0, 0)
```
DanilYachmenev, line 327: I'd actually like to see more tests for `pd.Timestamp` rather than for `datetime` since we use it more often

and also pls do a test case where `datetime` and `pd.Timestamp` will be endpoints of 1 interval

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +207,129 @@ def test6(self) -> None:
...
+
+class TestFindMinMaxTimestampsFromIntervals(hunitest.TestCase):
+    def test1(self) -> None:
+        """
+        Check for the case when no intervals are provided.
+        """
+        intervals = None
+        actual = dtfcorutil.find_min_max_timestamps_from_intervals(intervals)
+        expected = (None, None)
+        self.assertEqual(expected, actual)
```
DanilYachmenev, line 220: place `actual` before `expected` everywhere in the test

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +207,129 @@ def test6(self) -> None:
...
+        Check for an interval with different endpoint types.
+        """
+        intervals = [
+            (
+                datetime.datetime(2022, 1, 1, 10, 0, 0, tzinfo=pytz.utc),
+                pd.Timestamp(2022, 1, 1, 11, tz="UTC"),
+            )
+        ]
+        actual = dtfcorutil.find_min_max_timestamps_from_intervals(intervals)
+        expected_min_max = (
```
DanilYachmenev, line 331: just `expected`
maintain name consistency

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +207,129 @@ def test6(self) -> None:
...
+        )
+        self.assertEqual(expected, actual)
+
+    def test8(self) -> None:
+        """
+        Check for an interval with different endpoint types.
+        """
+        intervals = [
+            (
+                datetime.datetime(2022, 1, 1, 10, 0, 0, tzinfo=pytz.utc),
```
DanilYachmenev, line 326: do we really need `tz` here?
this also requires adding 1 more import
can we drop it just in this case?

dataflow/core/test/test_utils.py
```diff
@@ -201,3 +207,129 @@ def test6(self) -> None:
         actual_df = dtfcorutil.convert_to_multiindex(df, asset_id_col)
         # Compare the result.
         self.assert_multiindex_columns_equal(expected_df_columns, actual_df)
+
+
+class TestFindMinMaxTimestampsFromIntervals(hunitest.TestCase):
```
DanilYachmenev, line 212: -> `Test_find_min_max_timestamps_from_intervals(...`

DanilYachmenev (commented):
lgtm
fyi @samarth9008

samarth9008 (approved):
LGTM

PR #920, opened 2024-05-08
SorrTask830_PEOE_first_step (EDA)

tkpratardan:
I have done some preliminary data analysis on the cricket and football datasets. I am trying to understand how to interpret the formulation of betting odds and lines to be able to perform meaningful analysis on these features. 
#1010 
@gpsaggese @sonaalKant @samarth9008


gpsaggese (commented):
Going in the right direction.

1) Can you ask @samarth9008 how to use our env so we use jupytext?
2) let's create a dir research_amp/sports_prediction
3) let's collect results of the analysis in a gdoc

Issue #992, opened 2024-05-17
Initial steps on soccer dataset


sonaalKant:
Lets follow the steps to have experiments in organized manner, We will move the data to S3 so that every one can play with the notebook :
- Before proceeding with further EDA, lets divide the data in train/val/test
    - Use a fixed random seed( eg: 12345) so that split is reproducible every time. 
    - If you think the dataset size is small and we don't have enough test samples then do k-fold cross validation
- Do further data analysis / feature engineering on **train set.**
  - you can also start with feature importance ( how features are correlated with label)
 
- Document
   - Try to  maintain your findings / experiments done on the data ( failed or successful everything should be logged)  organized under the header date ( which will be the date you performed those experiments ).

FYI @samarth9008 @gpsaggese @Shaunak01

sonaalKant:
This is complete, @tkpratardan can you post the link of the PR where your work so that we can close.

tkpratardan:
It is PR #994.

PR #994, opened 2024-05-20
SorrTask992_initial_steps_on_soccer_dataset

tkpratardan:
@gpsaggese @sonaalKant @samarth9008 @Shaunak01 '

The files here try to establish a baseline system using some feature engineering techniques. #992

P.S: If we need to use pymc module, we have to change the docker image to have this module installed.


research_amp/sports_prediciton/baseline_system.py
```diff
@@ -0,0 +1,768 @@
+# ---
```
gpsaggese, line 1: let's move to soccer_prediction (there is a typo in the file name).

Then let's merge this and do small steps starting from SOTA.

gpsaggese (commented):
Good work overall.

We need to have a clear research plan, starting from SOTA, and then use some of the techniques we use for finance prediction, which is different than standard ML.

We'll talk about all this.

research_amp/sports_prediciton/baseline_system.py
tkpratardan: My bad. I completely missed it. :D

I will do the changes.

PR #1027, opened 2024-06-04
SorrTask1026_uptade_notebook

tkpratardan:
Updates to sota_baseline_glm.ipynb to analyze the goodness of fit.


tkpratardan:
Points:
- Contant training accuracy across different folds -> Does it suggest overfitting?
- High variation in validation accuracy -> again, does it suggest overfitting?

My suggestion: Can we begin experimenting with regularization?

research_amp/soccer_prediction/sota_baseline_glm.py
```diff
@@ -183,27 +185,88 @@ def representative_sample(df: pd.DataFrame(), sample_size: int) -> pd.DataFrame(
...
+    residuals = poisson_model.resid_deviance
+    plt.figure(figsize=(10, 6))
+    plt.hist(residuals, bins=30, edgecolor='k', alpha=0.7)
+    plt.title('Distribution of Residuals')
+    plt.xlabel('Residuals')
+    plt.ylabel('Frequency')
+    plt.show()
+    # Save the trained model. 
+    local_model_path = 'poisson_model.pkl'
+    with open(local_model_path, 'wb') as f:
```
gpsaggese, line 260: There is a function in jpickle

research_amp/soccer_prediction/sota_baseline_glm.py
```diff
@@ -183,27 +185,88 @@ def representative_sample(df: pd.DataFrame(), sample_size: int) -> pd.DataFrame(
     # Return the sampled dataframe.
     return sampled_df
 
-def train_model(train_df: pd.DataFrame(), 
+def train_model(train_df: pd.DataFrame,
```
gpsaggese, line 188: let's move all the code to a lib

gpsaggese:
Good work again.

Our research process suggests to
1) add a paragraph to the lab notes
You add the date, point to this PR, and then explain what you did and what you have observed.
You can copy paste plots, etc.

2) Then we can discuss the results in that gdoc

gpsaggese:
Since I'm super curious, I couldn't wait and I took a look at the notebook.

1) Small overfit. The variations are not too bad (we'll need conf intervals)
2) The model is really simple (even if it has a lot of params), so things are looking good
3) We should check the p-values of the model
4) The model is still optimistic since it's missing the rolling component, but it's ok for now we'll do this in KaizenFlow
5) I predict that the tree-based models will overfit like crazy out of the box. IMO this model complexity is the best with this amount of data
6) We need to add diagnostics to make sure the outputs are sensitive (e.g., do we get a decent hit rate for all the classes, etc)
- A measure of "better than naive" for every fold, etc.

The idea is to set up a framework to run experiments that is super robust, before starting complicating the model, otherwise we can't really decide if a model is a net improvement or not and we run in circles.

In any case, it's looking good.

PR #1033, opened 2024-06-10
SorrTask1029_mle_predictions

tkpratardan:
- Added MLE predictions
- Increased the training set to 40k samples from 20k samples
- Added functions to compute stats on input df (value counts of classes and prior probabilites)


research_amp/soccer_prediction/sota_baseline_glm.py
```diff
@@ -183,27 +201,88 @@ def representative_sample(df: pd.DataFrame(), sample_size: int) -> pd.DataFrame(
...
+    kf = sms.KFold(n_splits=n_splits, shuffle=True, random_state=42)
+    train_accuracies = []
+    val_accuracies = []
+    for fold, (train_index, val_index) in enumerate(kf.split(train_df), 1):
+        train_split = train_df.iloc[train_index]
+        val_split = train_df.iloc[val_index]
+        # Fit the Poisson regression model on the training split.
+        poisson_model = smf.glm(
+            formula=formula, data=train_split, family=sm.families.Poisson()
+        ).fit(maxiter=10)
```
sonaalKant, line 229: hyper param should be global.

research_amp/soccer_prediction/sota_baseline_glm.py
```diff
@@ -183,27 +201,88 @@ def representative_sample(df: pd.DataFrame(), sample_size: int) -> pd.DataFrame(
...
+    plt.ylabel('Coefficient Value')
+    plt.show()
+    # Plot the residuals.
+    residuals = poisson_model.resid_deviance
+    plt.figure(figsize=(10, 6))
+    plt.hist(residuals, bins=30, edgecolor='k', alpha=0.7)
+    plt.title('Distribution of Residuals')
+    plt.xlabel('Residuals')
+    plt.ylabel('Frequency')
+    plt.show()
```
sonaalKant, line 273: Similar to above.

research_amp/soccer_prediction/sota_baseline_glm.py
```diff
@@ -183,27 +201,88 @@ def representative_sample(df: pd.DataFrame(), sample_size: int) -> pd.DataFrame(
...
-            )
+    # Visualize the coefficients.
+    coeffs = poisson_model.params
+    plt.figure(figsize=(10, 6))
+    coeffs.plot(kind='bar')
+    plt.title('Model Coefficients')
+    plt.xlabel('Features')
+    plt.ylabel('Coefficient Value')
+    plt.show()
+    # Plot the residuals.
```
sonaalKant, line 266:
```
poisson_model.params.plot(kind='bar', figsize=(10, 6), title='Model Coefficients', xlabel='Features', ylabel='Coefficient Value', legend=False)
plt.show()
```

research_amp/soccer_prediction/sota_baseline_glm.py
```diff
@@ -183,27 +201,88 @@ def representative_sample(df: pd.DataFrame(), sample_size: int) -> pd.DataFrame(
...
+    # Save the trained model. 
+    local_model_path = 'poisson_model.pkl'
+    with open(local_model_path, 'wb') as f:
+        pickle.dump(final_model, f)
+    # Define the S3 bucket and path.
+    bucket_name = 'cryptokaizen-data-test'
+    s3_model_path = 'kaizen_ai/soccer_prediction/models/poisson_model.pkl'
+    # Upload the model to S3.
+    s3 = haws.get_service_resource(aws_profile = 'ck', service_name = 's3')
+    s3.Bucket(bucket_name).upload_file(local_model_path, s3_model_path)
```
sonaalKant, line 283: This should definitely be another method, which takes any model and filepath and saves the model in s3.

sonaalKant (changes requested):

research_amp/soccer_prediction/sota_baseline_glm.py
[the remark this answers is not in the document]
tkpratardan: > epsilon isn't too big?. Lets try to use all the hyperparams in config and outside functions so that the code is to tweak.

because we use logs in estimation, lower values of eplison implies higher values of log than that in the usual range ( no. of goals lies between 1-10)

gpsaggese:
Agree with @sonaalKant points.

The root cause is that we are writing too much code without thinking about the architecture / writing maintainable code.
The good news is that it's an easy problem to fix and we can help.

Let's focus on cleaning and re-organizing the code.

tkpratardan:
@sonaalKant @gpsaggese Closing this PR and the file #1036 is meant to cater to the changes mentioned here. PR #1049

Issue #1045, opened 2024-06-13
Add unit test for get_s3_bucket_from_stage


samarth9008:
Function - 

- Check all the corner cases are correctly handled by the function.

See unit test [doc](https://github.com/kaizen-ai/kaizenflow/blob/master/docs/coding/all.write_unit_tests.how_to_guide.md) to follow the code style
Also highly recommend to read this [doc](https://github.com/kaizen-ai/kaizenflow/blob/master/docs/coding/all.submit_code_for_review.how_to_guide.md#save-reviewer-time) before submitting 1st PR

FYI @gpsaggese @sonaalKant @DanilYachmenev

samarth9008:
@sonaalKant LMK once you update the link. I can re-route

PR #1051, opened 2024-06-14
Added Unit tests for get_s3_bucket_from_stage()

mayank922:
Task  #1045


mayank922:
Hi @samarth9008 

I have a few questions regarding how the optional suffix input is being provided to the function.
I'm assuming the suffix acts as a directory for the bucket 

https://github.com/kaizen-ai/kaizenflow/blob/249f93c1b8f70e6ad9fd18f054f9160f130894cb/helpers/hs3.py#L983

1. How would it be handled if I provided a string with spaces?  
`Example: add_suffix = "directory subdirectory"`
2. Can the suffix be an empty string?

helpers/test/test_hs3.py
```diff
@@ -346,3 +346,46 @@ def test2(self) -> None:
...
+class Test_get_s3_bucket_from_stage(hunitest.TestCase):
+    def test1(self) -> None:
+        """
+        Check for a valid stage.
+        """
+        # Define arguments
+        stage = "test"
+        # Run
+        hs3.get_s3_bucket_from_stage(stage)
+
```
gpsaggese, line 363: Can you add a check to make sure that the outcome is what is expected?
Otherwise you are just checking that it runs.

helpers/test/test_hs3.py
```diff
@@ -346,3 +346,46 @@ def test2(self) -> None:
...
+# #############################################################################
+
+
+class Test_get_s3_bucket_from_stage(hunitest.TestCase):
+    def test1(self) -> None:
+        """
+        Check for a valid stage.
+        """
+        # Define arguments
```
gpsaggese, line 359: Need a `.` after each comment

helpers/test/test_hs3.py
mayank922: To check that the function returns the expected output, should I use 
`self.assert_equal()`

samarth9008: Yes

mayank922:
Hi @samarth9008 

Are there any other issues to be resolved in this PR?

gpsaggese (commented):
LGTM

PR #1069, opened 2024-07-01
KaizenTask1050_improve_poisson_model_predictions

tkpratardan:
This is a draft PR of the model from the implementation of this [reference](https://journals.sagepub.com/doi/pdf/10.1177/1471082X18817650). 

The model uses the bivariate implementation and time weightage as suggested in the paper. However, conventional poison regressor from Stats models cannot be used in this way of implementation.


sonaalKant:
Closing this experiment branch as well.

Issue #1075, opened 2024-07-10
Unit test hpandas `dassert_valid_remap`


sonaalKant:
Lets test the method to show all assert cases are caught.

https://github.com/kaizen-ai/kaizenflow/blob/bc8bf878faa174a77c5d23fe7976de7307b1d717/helpers/hpandas.py#L247

Steps: 
1. Create multiple mock inputs and verify that all the dassert catches the invalid inputs correctly.
2. Follow the DRY principle to avoid writing repeated code.

FYI @samarth9008

PR #1080, opened 2024-07-13
SorrTask1075_Unit_test_hpandas_dassert_valid_remap

smitpatel49:
#1075


smitpatel49:
Can you please give me suggestions on how to improve on this one @sonaalKant, @samarth9008 , especially test3 and test4 as I was expecting an error raise but it went through. Thank you.

samarth9008:
Can you show the error while running test 3 and 4?

samarth9008:
> Also, what other input variations could be made, especially with to_remap for rows / columns remap.

I would check all the conditions, assertions where the function would break and check those conditions.

helpers/test/test_hpandas.py
```diff
@@ -3963,25 +4042,19 @@ def get_multiindex_df(
 
         Example of dataframe returned when `index_is_datetime = True`:
```
gpsaggese, line 4044: Revert this linter change.

Sometimes the linter wants an extra empty line to work properly.

gpsaggese (commented):
LGTM. One comment about checking the PR after the linter

helpers/test/test_hpandas.py
samarth9008: Revert any code changes that are messed up by linter

smitpatel49:
@samarth9008, I just wanted to make sure that my method in test 3 was correct and I didn't have any error in test4, but I was curious regarding the input as it had duplicate keys in the dictionary and was expecting some error. 
```
remap_dict = {
            "dummy_value_1": "1, 2, 3",
            "dummy_value_2": "A, B, C",
            "dummy_value_1": "A1, B2, C3",
        }
```

helpers/test/test_hpandas.py
smitpatel49: I will revert that linter change and will keep that in mind.

gpsaggese (commented):
LGTM

samarth9008 (commented):
Lets add tests for this 2 lines as well

```
hdbg.dassert_isinstance(to_remap, list)
hdbg.dassert_isinstance(remap_dict, dict)
```

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,83 @@ def test3(self) -> None:
...
+        """
+        # Set inputs.
+        to_remap = ["dummy_value_1", "dummy_value_2", "dummy_value_3"]
+        remap_dict = {
+            "dummy_value_1": "1, 2, 3",
+            "dummy_value_2": "A, B, C",
+            "dummy_value_1": "A1, B2, C3",
+        }
+        # Check.
+        hpandas.dassert_valid_remap(to_remap, remap_dict)
```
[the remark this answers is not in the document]
smitpatel49: Yes, that's what I thought. that python doesn't allow for duplicate keys as it is but the function explicitly mentions for no duplications of keys and values so I was confused. I will create an issue on this as you mentioned on this soon. 
Also, you mean [this file](https://github.com/kaizen-ai/kaizenflow/blob/65fc4bde2c2854482672559f6c5cb0bfcd916f67/helpers/hdbg.py) because the link you mentioned does not work.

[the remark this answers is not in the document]
samarth9008: > Also, you mean [this file](https://github.com/kaizen-ai/kaizenflow/blob/65fc4bde2c2854482672559f6c5cb0bfcd916f67/helpers/hdbg.py) because the link you mentioned does not work

Yes this file

smitpatel49:
Awaiting further review/ instructions on this one. @samarth9008, @gpsaggese,  @sonaalKant , @DanilYachmenev

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,109 @@ def test3(self) -> None:
 # #############################################################################
 
 
+class Test_dassert_valid_remap(hunitest.TestCase):
+    def test1(self) -> None:
+        """
+        A simple test to check that the function works.
```
samarth9008, line 141: Check that the function works with correct inputs.

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,109 @@ def test3(self) -> None:
...
+        remap_dict = {
+            "dummy_value_1": "1, 2, 3",
+            "dummy_value_2": "A, B, C",
+        }
+        # Check.
+        hpandas.dassert_valid_remap(to_remap, remap_dict)
+
+    def test2(self) -> None:
+        """
+        Check that an assert is raised if dictionary keys are not a subset.
```
samarth9008, line 154: assert -> assertion

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,109 @@ def test3(self) -> None:
...
+        """
+        # Set inputs.
+        to_remap = ["dummy_value_1", "dummy_value_2"]
+        remap_dict = {
+            "dummy_value_1": "1, 2, 3",
+            "dummy_value_2": "A, B, C",
+            "dummy_value_3": "A1, A2, A3",
+        }
+        # Run
```
samarth9008, line 163:
```
# Run.
```

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,109 @@ def test3(self) -> None:
...
+        issubset
+        val2=['dummy_value_1', 'dummy_value_2']
+        val1 - val2=['dummy_value_3']
+        Keys to remap should be a subset of existing columns"""
+        # Check.
+        self.assert_equal(actual, expected, fuzzy_match=True)
+
+    def test3(self) -> None:
+        """
+        Check if duplicate values create error.
```
samarth9008, line 179: Check that an assertion is raised if the duplicate values are present in the dict.

Keep the style of comments consistent.

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,109 @@ def test3(self) -> None:
...
+            "dummy_value_1": "1, 2, 3",
+            "dummy_value_2": "A, B, C",
+            "dummy_value_3": "1, 2, 3",
+        }
+        # Run
+        with self.assertRaises(AttributeError) as cm:
+            hpandas.dassert_valid_remap(to_remap, remap_dict)
+        actual = str(cm.exception)
+        expected = r"""
+        'dict_values' object has no attribute 'count'"""
```
samarth9008, line 193: The assertion should be something like this which represents duplicate values

```
txt = []
        txt.append("val1=\n" + pprint.pformat(val1))
        txt.append("has duplicates")
        txt.append(",".join(map(str, dups)))
        _dfatal(txt, msg, *args, only_warning=only_warning)
```

The current one is related to count. Can you pass single int values as value in the dict?

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,109 @@ def test3(self) -> None:
...
+        # Check.
+        self.assert_equal(actual, expected, fuzzy_match=True)
+
+    def test4(self) -> None:
+        """
+        Check if an error is raised when the instance is not a list.
```
samarth9008, line 199: Change the comment as per above review. Also, what instance are talking about? Instead its better to define what should be list and what should not be.

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,109 @@ def test3(self) -> None:
...
+        Check if duplicate values create error.
+        """
+        # Set inputs.
+        to_remap = ["dummy_value_1", "dummy_value_2", "dummy_value_3"]
+        remap_dict = {
+            "dummy_value_1": "1, 2, 3",
+            "dummy_value_2": "A, B, C",
+            "dummy_value_3": "1, 2, 3",
+        }
+        # Run
```
samarth9008, line 188: -> `# Run.`


Check at other places as well.

helpers/test/test_hpandas.py
```diff
@@ -134,6 +135,109 @@ def test3(self) -> None:
...
+        expected = r"""
+        * Failed assertion *
+        Instance of '{'dummy_value_1'}' is '<class 'set'>' instead of '<class 'list'>'
+        """
+        # Check.
+        self.assert_equal(actual, expected, fuzzy_match=True)
+
+    def test5(self) -> None:
+        """
+        Check if an error is raised when the instance is not a dictionary.
```
samarth9008, line 219: Change the comment as per above review.

samarth9008 (commented):
Some nits from my side.

helpers/test/test_hpandas.py
smitpatel49: I need to talk to you regarding test3; I have made other changes as you suggested. If you can have some time today, I would like to talk over the phone on this one. Thanks.

helpers/test/test_hpandas.py
smitpatel49: I need to talk to you regarding test3; I have made other changes as you suggested. If you can have some time today, I would like to talk over the phone on this one. Thanks.

samarth9008: Can you debug and provide a reason why it is giving that assertion and not related to duplicate values itself. My guess is, it might have to do something with how dictionary are handled in python. 


If the provided reason looks valid, the PR looks good.

smitpatel49: I don't think its the way python is handling dictionaries because python allows duplicate values. The reason it throws error is due to:
`def dassert_no_duplicates(
    val1: Any, msg: Optional[str] = None, *args: Any, only_warning: bool = False
) -> None:
    cond = len(set(val1)) == len(val1)
    if not cond:
        # Count the occurrences of each element of the seq.
        v_to_num = [(v, val1.count(v)) for v in set(val1)]`

So, the function `dassert_no_duplicates` shouldn't be used to check duplicates for a dictionary in general.

smitpatel49: I don't think its the way python is handling dictionaries because python allows duplicate values. The reason it throws error is due to:

```
`def dassert_no_duplicates(
    val1: Any, msg: Optional[str] = None, *args: Any, only_warning: bool = False
) -> None:
    cond = len(set(val1)) == len(val1)
    if not cond:
        # Count the occurrences of each element of the seq.
        v_to_num = [(v, val1.count(v)) for v in set(val1)]`
```

So, the function `dassert_no_duplicates` shouldn't be used to check duplicates for a dictionary in general.

samarth9008: how about if we pass list of values?

smitpatel49: I think it will throw a TypeError because we would be dealing with nested lists.

samarth9008 (commented):
Did we add check for this LOC -  [hdbg.dassert_not_intersection(remap_dict.values(), to_remap)](https://github.com/kaizen-ai/kaizenflow/blob/bc8bf878faa174a77c5d23fe7976de7307b1d717/helpers/hpandas.py#L263C4-L263C65)

smitpatel49:
That was my initial question regarding the input, the function mentions row/ columns to map, the `to_remap` list will contain columns as in keys but how to include values as when calling `hdbg.dassert_is_subset` we take the list as whole - just containing keys.

gpsaggese:
What's the next step here? IMO it's best to merge and do a follow up PR. We don't want PRs on deck for 2 weeks. A PR should be merged in 1-2 days max.

smitpatel49:
I think that would be appropriate. I have been waiting on review from others and active response for a very long time on this PR.

samarth9008 (approved):
LG merging.

samarth9008:
> We don't want PRs on deck for 2 weeks. A PR should be merged in 1-2 days max.

The blockage was from my side as I have multiple things going on so couldn't find time to review these PRs.

